PR #7, opened 2018-01-12
Merge initial work into official initial commit

dpordomingo:


src/logo.svg
```diff
@@ -0,0 +1,7 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 841.9 595.3">
```
bzz: From https://github.com/facebook/react/blob/master/LICENSE this logo seems to be MIT licensed, which is compatible with GPL3 but we would need to track this somewhere, if we start mixing licenses inside a single codebase.

I would suggest we either remove it, or replace it with something we own, that can be covered by GPL 3 as everything else in this repo.

bzz (changes requested):

bzz:
Looks great to me, sans minor potential licensing issue noted above.

dpordomingo:
Thanks @bzz for reviewing the licenses.
@smacker since everything is yours, but the last commit 4c7b9e9dbe9349168829359d6b2f6b6289a2468c that's the only one that you'd review :D

dpordomingo:
PS: @bzz should we remove that icon from everywhere in our history? or is it enough with that commit 4c7b9e9 removing it? If so, I can do it.

bzz:
It's enough not to have it in the version of the source code, that will be distributed. 

It's actually preferable to have it in the history with excellent commit message on how it's been taken care of, before it became an issue.

Thank you @dpordomingo for taking care of it - looks great to me.

bzz (approved):
LGTM 

Let's merge, if there is no further discussion so it could serve as excellent foundation for future work!

Issue #1, opened 2018-01-12
Initial frontend work


bzz:
This task is about setting up pages  from Informational Architecture section of [Design Doc](https://docs.google.com/document/d/1zxBaQRmt-mZ5H0P0X4OK65QPJfA5UQ9wb3vLZ9NCd_o/edit?userstoinvite=rizome.es@gmail.com&ts=5a567b81&actionButton=1#) and components to display 2 files with the diff highlighted.

Issue #3, opened 2018-01-12
Backend: REST API


bzz:
basics 
- [x] prepare skeleton with handlers returning `http.StatusNotImplemented`
- [x] handle GitHub authorization
- [x] implement handlers, returning mocked data
- [x] using DB stuff https://github.com/src-d/code-annotation/issues/41
- [ ] connect front with api #56 
- [ ] end2end tests for the API ??

nice to have
- [ ] log 500 error https://github.com/src-d/code-annotation/issues/50
- [ ] restrict access of unknown users https://github.com/src-d/code-annotation/issues/48

smacker:
```
GET /experiments/:id
{
  data: {
    id: <id>,
    name: TEXT,
    description: TEXT
  }
}

GET /experiments/:id/assignments
{
  // should we filter only by assignments with answer? if so we need to return total
  // somewhere and what to do with id? actually for simplicity I would return everything.
  // 1k isn't that much.
  data: [{id, pairId, experimentId, answer, duration}, ...]
}

answer: Yes, No, Maybe, Skip

PUT /experiments/:id/assignments/:id
{
  answer,
  duration,
}

GET /experiments/:id/filePairs/:id
{
  data: {
    id,
    diff
  }
}
```

dpordomingo:
Some considerations:
(@bzz tell me if we should somehow persist this in the _DesignDocument_)

### auth:
Auth with Github, auth communication between front and back using [JWT](https://jwt.io/introduction) and [dgrijalva/jwt-go](https://github.com/dgrijalva/jwt-go)

### experiments listing:
```
GET /experiments/:id/assignments
```
- `experimentId` wont be sent
- it will return the assignments for the user identified by the `HEADER` user token 
- it will create all user assignments when it is requested for the first time

Since all assignments are sent at once to the front, all metrics and aggregations will be done in the browser.

### sending answers:
The browser will send to the backend the answers whenever it wants (it can be onAnswer, onNext, onClose, onFinish, async...).

When it is sent an assignment, the backend will override the stored one with the data sent, so the frontend will consider if the duration is reset, or added to the previous one.

The frontend will take care of race conditions, many tabs sending the same answer or sending it from the same tab many times in a short period of time. 

 If @smacker wants, the "save" event could be done "all at once", with, for example:
```
PUT /experiments/:id/assignments
[{
  id,
  answer,
  duration,
},...]
```

### errors:
[...]

PR #9, opened 2018-01-15
Initial frontend

smacker:
Fixes https://github.com/src-d/code-annotation/issues/1


bzz:
Looks ✨ awesome ✨ !

Regarding the questions above.

1. On routing

Indeed, it would be nice to have only what we need, but I think we might need routing 
 - for annotations: so a link to particular pari of files could be shared (i.e in case there are questions about it)
 - for reviewers: we need to be able to separate it from annotations and have 2 views
   * for pair of files - overall stats of user replies
   * for same pair of files - answer of each particular user

It seems to me that routing is the most natural way of separating those views and keeping the ability of having each of them to have a uniq url. Or do you have any other suggestions in mind?

2. On tests.

Up to you, but I believe that having at least minimum amount of tests for things that are most error prone would be is a good idea. I.e this would show further contributors how to do testing and might be a safe net for future refactoring.

3. > how much logic are we going to keep on FE and what is going to be on BE? For example, will server return only assignments or some meta information like statistics too?

Could you elaborate, what statistics do you mean? In case of reviewing (case 5 from Information Architecture section of [DesignDoc](https://docs.google.com/document/d/1zxBaQRmt-mZ5H0P0X4OK65QPJfA5UQ9wb3vLZ9NCd_o/edit#heading=h.4dye7cd2fhl0)), I would expect backend to provide at least:
  - pairs of files
  - their diff
  - ML "features" to display
  - ML similarity "score"
  - histogram of answers

4. > can we define API endpoints

I believe it will be done under https://github.com/src-d/code-annotation/issues/3 but it seems to be neither assigned, nor have any updates of the progress. Feels like at this point you can take advantage of this and post any suggestions :)

smacker:
Thank you for your answers Alex!

1. All your points definitely make sense. I just wasn't sure how sharing it is important to us. Looks like it is important. I'll implement routing.

2. I would like to add tests when I integrate it with API. In my experience real api can be quite different from what I expect. And sometimes it requires changing even view components. So the plan is:
1) wait for BE
2) use API
3) test everything manually (need to do anyway)
4) write tests, to avoid mistakes in future

Hope it's alright.

3. I meant something like how many times user answered `similar`, how many `different`, how much time he/she spent and so on. I'm fine with calculating it on FE btw. So it's more about do guys on BE want to implement it there or not.

4. Issue is more about implementation. I'm talking about a contract. So I could start implementing api requests/data layer in parallel with work on BE.

bzz:
All sounds great! 

on 2. Tests
We have a section on Testing in Design Doc that I have updated to reflect content of this conversation.

on 3. Stats for review
I would expect this to be calculated on BE to avoid sending all datapoints at once to FE for potentially big experiments. May be a separate issue would need to be created for that work (does not matter, BE/FE).

on 4. REST API contract
I would expect having endpoints/message structure defined as The first step of #3

smacker:
Guys, please also let me know if you feel the code should be commented. I feel it is self-describing enough without comments in the current state, but I can be wrong. 😅

src/pages/Index.js
```diff
@@ -0,0 +1,27 @@
...
+import { connect } from 'react-redux';
+import { logIn } from '../state/user';
+import './Index.less';
+
+class Index extends Component {
+  render() {
+    const { logIn: onClick } = this.props;
+    return (
+      <div className="index-page">
+        <h1>Welcome! We&#39;re glad you make it this far.</h1>
```
carlosms: Should be "you made it" / "you've made it"?

carlosms (approved):
LGTM

smacker:
I added a router. It's hard as usual 😞

src/pages/Index.js
smacker: yes. I just copied it from wireframes. Fixed.

bzz:
Very nice!

One thing  - I would suggest everyone to setup commit sign-off \w DCO to comply with [src-d contributing guideline](https://github.com/src-d/guide/blob/master/engineering/documents/CONTRIBUTING.md#certificate-of-origin).

After doing this, something like `git rebase --exec 'git commit --amend --no-edit -n -S' -i HEAD~<N>` would sign-off last N commits for you.

bzz:
Very nice! 

For some reason, I have mixed [commit signing](https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work) or `-s` and sign-off \w the DCO `-S`.

vmarkovtsev:
I request a staging/playground link for people who are not sure how to setup the development environment, including myself.

smacker:
@vmarkovtsev it isn't ready yet for staging. But it's easy to run locally.
```
git clone
make serve
```
as described in README.

You need only go, node and yarn installed locally.

vmarkovtsev:
@smacker ML team does not know anything about node and yarn and yet we want to play. Even installing node in our Ubuntu-s is not trivial, with all that `nodejs-legacy` package shit. Besides,

```
vadim@vadim-XPS-13-9360:~/Projects/code-annotation$ make serve
go get -v -t _/home/vadim/Projects/code-annotation/server/cmd/code-annotation
package _/home/vadim/Projects/code-annotation/server/cmd/code-annotation: unrecognized import path "_/home/vadim/Projects/code-annotation/server/cmd/code-annotation" (import path does not begin with hostname)
Makefile.main:82: recipe for target '_/home/vadim/Projects/code-annotation/server/cmd/code-annotation' failed
make: *** [_/home/vadim/Projects/code-annotation/server/cmd/code-annotation] Error 1
```

Do not expect ML team to be able to build anything, we really need smth already running somewhere or a Docker container.

smacker:
@vmarkovtsev we will have staging&docker in the future for sure. I just suggested to play very quickly because it gives you better experience than wireframes. If it's too difficult to run, forget about it... or you can come over and take a quick look on my laptop :)

vmarkovtsev:
Created https://github.com/src-d/code-annotation/issues/11

smacker:
@bzz @dpordomingo @carlosms guys, I suggest to review & merge on this point.

It works somehow (without backend), it looks similar to current wireframes, it has some tests and so on.
It's a good foundation to start with.

After merge we can create more detailed issues and fix them in parallel.

vmarkovtsev:
We should really speed up merging this - we need to show this to possible external users already

bzz:
@vmarkovtsev what would help to speed things except for code reviews is a report that one was able to successfully run it locally :)

smacker:
@bzz Carlos run it locally successfully. He reported it on slack yesterday

bzz:
@smacker it's good to know! It would make sense to keep it posted here.

Let me get back to you on this asap.

bzz:
After running `make serve` user gets a message 

```
  "homepage" : "http://myname.github.io/myapp",
```
May be I'm doing something wrong, so let me know.

@smacker would it be possible to display the URL that user can click on to see the app?

Upd: documented way of `yarn start` opens a browser window for the user 🎉

bzz:
Not sure if that is a 🐛 or expected behavior, but after

 - sign in
 - skip the question
 - check the checkbox "always skip"
 - finish annotations
 - sign in again

and one has a modal dialog of skipping the question right away, on the first question, without clicking anything, just after being re-directed to the `/exp/1/3`.

Need @smacker feedback to triage this.

src/state/experiment.js
```diff
@@ -0,0 +1,266 @@
+import { createSelector } from 'reselect';
+import { LOCATION_CHANGED, replace, push } from 'redux-little-router';
+import api from '../api';
+import { namedRoutes, makeUrl } from './routes';
+import { add as addErrors } from './errors';
+
+export const ANSWER_SIMILAR = 'Yes';
+export const ANSWER_MAYBE = 'Maybe';
+export const ANSWER_DIFFERENT = 'No';
+export const ANSWER_SKIP = 'Skip';
```
carlosms, line 10: Minor comment on consistency: Buttons are labeled Identical, Similar, Different, while the drop-down is Yes, Maybe, No

smacker: I'll change them as soon as API will be ready. I mean in another task about API integration.

smacker:
@bzz 
1) `make serve` runs `yarn build`. Which runs default build script of create-react-app. And they decided to include tip there.
Anyway it isn't much relative. For `make serve` URL must be generated by backend. 
2) popup was removed from wireframes. We will remove it completely in another task.

bzz (approved):
LGTM

src/state/experiment.js
bzz: can be handled in https://github.com/src-d/code-annotation/issues/17

vmarkovtsev:
Thanks a lot guys for writing such a cool app in such a short time!

package.json
```diff
@@ -4,27 +4,39 @@
   "private": true,
   "dependencies": {
     "bootstrap": "3",
+    "diff2html": "^2.3.3",
+    "less": "^2.7.3",
```
dpordomingo, line 8: I understood we are using `sass`, as done in most of our web apps.
Is there any limitation about it?

src/api/index.js
```diff
@@ -0,0 +1,92 @@
+import mocks from './mocks';
+
+const defaultServerUrl =
+  process.env.REACT_APP_SERVER_URL || 'http://127.0.0.1:8080/api';
```
dpordomingo, line 4: Do we really need to serve the api under `/api` path?

src/api/index.js
```diff
@@ -0,0 +1,92 @@
...
+  if (resp.status < 200 || resp.status >= 300) {
+    const error = new Error(resp.statusText);
+    error.response = resp;
+    throw error;
+  }
+  return resp;
+}
+
+function normalizeError(err) {
+  if (typeof err === 'object') {
```
dpordomingo, line 18: `normalizeError` seems to process different kinds of errors, but some of them are not properly shown (it is only displayed "Internal error").
What if we use `switch (typeof err)`? and then:
- properly define the acceptable types,
- gracefully display the `default:` kinds, instead of hiding them.

src/components/Errors.less
```diff
@@ -0,0 +1,8 @@
+.errors {
+    position: absolute;
+    top: 20px;
+    left: 50%;
+    width: 600px;
```
dpordomingo, line 5: why not using a common with for all dialogs/alerts...?
Or even better, why not using a fixed dimension, as already defined by the Bootstrap grid system

src/components/Experiment/Footer.less
```diff
@@ -0,0 +1,28 @@
...
+
+        &:before,
+        &:after {
+            display: none;
+        }
+    }
+
+    &__actions {
+        .btn {
+            margin: 0 10px;
```
dpordomingo, line 19: I think it should not be needed to satiate the buttons.
`<ButtonToolbar>` from bootstrap already does it for free ;)

src/components/Experiment/Footer.less
```diff
@@ -0,0 +1,28 @@
...
+        }
+    }
+
+    &__buttons {
+        .btn {
+            margin-left: 10px;
```
dpordomingo, line 25: _sic_

src/components/Experiment/Footer.js
```diff
@@ -0,0 +1,50 @@
...
+  value,
+  select,
+  markSimilar,
+  markMaybe,
+  markDifferent,
+  skip,
+  finish,
+}) {
+  return (
+    <div className="ex-footer">
```
dpordomingo, line 16: we should use `<Grid><Row><Column /></Row></Grid>`

src/pages/Experiment.js
```diff
@@ -0,0 +1,216 @@
...
+      window.localStorage.removeItem(ALWAYS_ALLOW_SKIP_KEY);
+    }
+  }
+
+  render() {
+    const { user } = this.props;
+
+    return (
+      <div className="ex-page">
+        <PageHeader {...user} />
```
dpordomingo, line 60: In [our wireframes](https://github.com/src-d/product/files/1631705/wireframe-source-code-annotation.pdf) from https://github.com/src-d/product/issues/253, the app title and user avatar are pulled to the window borders, but they're not in this implementation
image

src/components/Experiment/Footer.js
dpordomingo, line 16: In [our wireframes](https://github.com/src-d/product/files/1631705/wireframe-source-code-annotation.pdf) from https://github.com/src-d/product/issues/253, the prev dropdown and the finish/skip buttons are pulled to the window borders, but they're not in this implementation

image

src/components/Experiment/Footer.js
```diff
@@ -0,0 +1,50 @@
...
+      <div className="container">
+        <div className="ex-footer__pager">
+          Previous:{' '}
+          <select value={value} onChange={e => select(e.target.value)}>
```
dpordomingo, line 20: "Previous" dropdown is not styled
image
We should maybe use `<DropdownButton />`

src/pages/Experiment.js
```diff
@@ -0,0 +1,216 @@
...
+      <Modal show={this.state.showModal} onHide={this.hideModal}>
+        <Modal.Header closeButton>
+          <Modal.Title>
+            Are you sure you want to skip this question?
+          </Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
+          <p>You can always come back to this question later.</p>
+          <label>
+            <input
```
dpordomingo, line 148: "Checkbox" is not styled, we should maybe use `<Checkbox />`
image

src/components/Experiment/Progress.less
```diff
@@ -0,0 +1,33 @@
...
+
+.ex-progress {
+    display: flex;
+
+    &__percent {
+        width: 35px;
+    }
+
+    &__bar {
+        width: 150px;
```
dpordomingo, line 11: we should use a fixed dimension, as already defined by the Bootstrap grid system

src/index.js
```diff
@@ -1,22 +1,30 @@
...
 import { Provider } from 'react-redux';
-import { createStore, applyMiddleware } from 'redux';
+import { compose, createStore, applyMiddleware } from 'redux';
+import { initializeCurrentLocation } from 'redux-little-router';
 import { composeWithDevTools } from 'redux-devtools-extension';
 import thunk from 'redux-thunk';
-import rootReducer from './state';
+import rootReducer, { middlewares } from './state';
+import { enhancer as routerEnhancer } from './state/routes';
 import './index.css';
```
dpordomingo, line 12: Why not `.less` as all our stylesheets?

src/pages/Experiment.less
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    &__main {
+        flex-grow: 1;
+        display: flex;
+        flex-direction: column;
+    }
+
+    &__progress {
+        margin-bottom: 20px;
```
dpordomingo, line 23: If we could rely on Bootstrap grid system, we could get rid of hardcoded spacings

src/pages/Final.less
```diff
@@ -0,0 +1,15 @@
...
+    p {
+        text-align: center;
+    }
+
+    p {
+        margin-bottom: 20px;
+    }
+
+    &__table {
+        max-width: 600px;
```
dpordomingo, line 12: We should use predefined sizes as defined by Bootstrap grid system, shouldn't we?

src/pages/Final.less
```diff
@@ -0,0 +1,15 @@
+.final-page {
+    h1,
+    p {
+        text-align: center;
+    }
+
+    p {
+        margin-bottom: 20px;
```
dpordomingo, line 8: Why not using Bootstrap defaults?

src/pages/Index.less
```diff
@@ -0,0 +1,52 @@
...
+    &__header,
+    &__description,
+    &__preview,
+    &__actions {
+        text-align: center;
+    }
+
+    &__header {
+        color: #ccc;
+        margin: 0 0 20px;
```
dpordomingo, line 18: Why not using Bootstrap defaults?

src/state/experiment.js
dpordomingo: Dropdown labels and buttons texts should not depend on the API but totally decoupled.

dpordomingo (commented):
Wow! you did an awesome work in a short period of time! Congrats.
Now the PR is already merged into `master`, this pending issues should be addressed separately.

package.json
smacker: bootstrap

src/api/index.js
smacker: yes

src/api/index.js
smacker: This function tries to make an error for human. If it isn't possible "Internal error" is the best option.

src/components/Errors.less
smacker: don't get you. this css shows error on top, not in the center of the screen. it's much more user friendly.

src/components/Experiment/Footer.js
smacker: why?

smacker: wireframes aren't finished or approved yet. We will have issue to update UI when we get approve from product team. Before it in my opinion any UI changes is just a waste of time

src/components/Experiment/Footer.js
smacker: it's up to Ricardo. I personally like native dropdowns.

src/components/Experiment/Footer.less
smacker: 1. up to Ricardo
2. ButtonToolbar is actually to make toolbar without spaces. We do have spaces in wireframes at least we had when I wrote this css

src/components/Experiment/Footer.less
smacker: what "sic" means?

src/components/Experiment/Progress.less
smacker: why

src/index.js
smacker: it doesn't use any less features.

src/pages/Experiment.js
smacker: again. Up to Ricardo. Also if you make your browser smaller it will be much closer to borders.

src/pages/Experiment.js
smacker: it will be removed completely

src/pages/Final.less
smacker: it doesn't look good

src/pages/Final.less
smacker: I don't think so.

src/pages/Index.less
smacker: why do?

src/state/experiment.js
smacker: what? sure they are decoupled. From my understanding, the idea of @carlosms is to reconsider constants values because they don't reflect what we are actually doing. Right now is the best time for it.

src/api/index.js
dpordomingo: What if it's passed a number, for example?
```js
Internal error.
Content received: 2
```
Would be more detailed.

src/components/Errors.less
dpordomingo: But we could use common wides

src/components/Experiment/Footer.js
dpordomingo: For the shake of the consistency, because we are doing a grid-like layout, and using the Bootstrap primitives would  make the layout more consistent.

dpordomingo: But we agreed on try to follow the wireframe, didn't we?

src/components/Experiment/Footer.js
dpordomingo: But we decided to use Bootstrap...

src/components/Experiment/Footer.less
dpordomingo: Wrong latin said.
It should be _idem_, an I was talking about https://github.com/src-d/code-annotation/pull/9#discussion_r162454882

src/components/Experiment/Progress.less
dpordomingo: For consistency

src/index.js
dpordomingo: [src/components/Errors.less](https://github.com/src-d/code-annotation/blob/master/src/components/Errors.less) neither, but for the shake of the consistency it's a good idea use the same stylesheets format everywhere.

src/pages/Experiment.js
dpordomingo: But we agreed on try to follow the wireframe, didn't we?

src/pages/Experiment.js
dpordomingo: Why? If so, why did we even merged it?

src/pages/Final.less
dpordomingo: I understood we shouldn't take design considerations beyond the Bootstrap defaults. Otherwise, further design iterations will be way more difficult to achieve.

dpordomingo:
I understand they're somehow coupled to the API since you mentioned it will be fixed when the API is ready.
https://github.com/src-d/code-annotation/pull/9#discussion_r162379976
If the dropdown labels shows the values sent/recieved from the API, it's wrong: there should be a mapping between the strings shown in the UI, and the values handled by the API, because the API could work with non human constants, and the UI should use user friendly wording.

src/state/experiment.js
dpordomingo: I understand they're somehow coupled to the API since you mentioned it will be fixed when the API is ready. https://github.com/src-d/code-annotation/pull/9#discussion_r162379976
If the dropdown labels show the values sent/recieved from the API, it's wrong: there should be a mapping between the strings shown in the UI, and the values handled by the API, because the API could work with non human constants, and the UI should use user friendly wording.

src/pages/Index.less
dpordomingo: I understood we shouldn't take design considerations beyond the Bootstrap defaults. Otherwise, further design iterations will be way more difficult to achieve.

package.json
dpordomingo: How come?
I understood we're using Bootstrap through compiled `bootstrap/dist/css/bootstrap.css` and `react-bootstrap` so it should not collide with our preprocessors, does it?

I also tried to use Sass in this project, and it seems to work as you can see at https://github.com/dpordomingo/code-annotation/pull/1/files

src/state/experiment.js
smacker: read code. they aren't coupled.

package.json
smacker: please read code. I import bootstrap variables. Later I may also need to import bootstrap functions. The agreement was to use the same preprocessor as a component library does. Because bootstrap4 isn't ready to use with react yet, we use less.

src/api/index.js
smacker: why do we need it?

src/components/Errors.less
smacker: why?

src/components/Experiment/Footer.js
smacker: why do you think we do grid-like layout?

src/state/experiment.js
dpordomingo: So... since the API does not affect these labels, there is no reason because dropdown and buttons currently use different labels, there is?

src/components/Experiment/Footer.js
smacker: it does follow as much as it make sense in my opinion for current development stage. Everything else related to differences in UI will be fixed here: https://github.com/src-d/code-annotation/issues/19

src/components/Experiment/Footer.js
smacker: and?

src/components/Experiment/Progress.less
smacker: I disagree. If you want to use bootstrap grid system, please provide normal explanation why. Like a, b, c is better than what I do right now. I have very strong arguments NOT use bootstrap grid system

src/pages/Experiment.js
smacker: wireframes don't show how the page should look at different resolutions.

src/pages/Experiment.js
smacker: 1) It got removed from wireframes.
2) Why not?

src/api/index.js
dpordomingo: I think it improves the clarity &rarr; dev experience

src/components/Errors.less
dpordomingo: For the shake of the consistency

src/components/Experiment/Footer.js
dpordomingo: because we want to be benefited by [the reasons because a grid based approach would improve our UI](https://www.canva.com/learn/grid-design/)

src/components/Experiment/Footer.js
dpordomingo: I understood we wanted to be benefited &mdash;if possible&mdash; from all Bootstrap elements...
@ricardobaeta could you clarify if we are designing our custom inputs/dropdowns/checkboxes... or are we trying to use the Bootstrap ones?

src/components/Experiment/Footer.less
dpordomingo: How come?
I just wrapped the `<Button>` inside a `<ButtonToolbar>` and that's what I got:

image


I also found it docummented in [react-bootstrap/docs/buttons](https://react-bootstrap.github.io/components/buttons/)

src/pages/Experiment.js
dpordomingo: I understood that the provided wireframes represented the most common case, not the corner cases... but I understand that I may be wrong...

src/api/index.js
smacker: for dev you have `console.log`. End user doesn't need to know details. I'm sure @ricardobaeta can explain it better. But there are many resources on the internet with best practices from UX for error handling. None of them says you should show a user a traceback or errors that don't make sense for the user.

src/components/Experiment/Footer.js
smacker: I don't. Proper flexbox or css-grid solves this problem much much better. I'm talking about grid as about weird bootstrap css rules, not as a design pattern.

src/components/Experiment/Footer.js
smacker: Bootstrap doesn't have select component. And I can understand why. It's very difficult to create proper custom select component. `dropdown button` is very different from `select` in user experience.

src/pages/Experiment.js
smacker: yes. that's correct. For corner cases we have in-browser testing with @ricardobaeta

src/components/Experiment/Footer.less
smacker: Oh. I thought you are talking about this one: https://getbootstrap.com/docs/3.3/components/#btn-groups-toolbar

I'm ok with using that toolbar if margins are the same as we want them to be.

src/pages/Experiment.js
dpordomingo: That's right, this modal was removed in last wireframe https://balsamiq.cloud/symld/p81qk/r72FF

src/api/index.js
dpordomingo: I'm not proposing to show such errors to the user.

The `normalizeError` is used internally to extract error messages, and it should be up to the UI components, to display them or not, or how to so.
Anyhow, `normalizeError` is already returning ["internal/unnexpected" errors](https://github.com/src-d/code-annotation/pull/9/files#diff-284817f1b6aad47c0ac6d61bd72f474cR27) as I'm talking about, and I think it's ok (but not done in a consistent way)

src/components/Experiment/Footer.js
dpordomingo: It does. There is a `FormControl` for form `select` (although imho, it makes more sense a `<FormControl>` for our purpose); [they recommends to use `FormControl` for `select`](https://react-bootstrap.github.io/components/forms/#forms-controls): 
> _Use &lt;FormControl> for &lt;input>, &lt;textarea>, and &lt;select>_

It seems to fit better with our current look and feel, doesn't it?

image

src/pages/Experiment.js
dpordomingo: So, if we agree that wireframes do not reflect the corner cases, we could fairly assume that title/paginator and avatar/progress/skip are pulled to the screen borders, couldn't we?

smacker: If there is no specification it's up to developer for initial implementation. Then it gets validated by Ricardo. Sorry, I'm not going to change anything just because another developer has different "vision". We have very solid process for design: https://github.com/src-d/guide/pull/100 I'm going to follow it to archive the best UI experience.

src/components/Experiment/Footer.js
smacker: https://getbootstrap.com/docs/3.3/components/#input-groups
`Avoid using <select> elements here as they cannot be fully styled in WebKit browsers.`
so actually react-bootstrap and bootstrap itself have different recommendation :-D

I'm ok to add `.form-control` class to select (by using `FormControl`)
It or something better will be done in #19 anyway.

dpordomingo: Both agree on one thing: avoid using `<select>`;
if we proceed so, the alternatives seem to be:
- follow react-bootstrap guidelines, and to use `<FormControl>`,
- follow both guidelines, and to use an alternative, for example `<DropdownButton>`.

smacker: Before I used https://github.com/JedWatson/react-select it's a well-done library. Just as an option. `<DropdownButton>` has quite different behavior. It's good only for small list of options. Not 1k as we use here.

dpordomingo: Let's see what @ricardobaeta says about having a dropdown/select/whatever with >100 items.
Anyhow, I'm glad on agreeing we're removing the `<select>` thing

Issue #4, opened 2018-01-12
Tool for Importing pairs of files


bzz:
A tool that can be called from CLI and an HTTP handler, that consumes sqlite DB with pairs of files (any schema for now, will be changed later) from FS and saves to the DDL from #2.

@EgorBu will eventually provide more details on the input schema.

bzz:
Based on https://github.com/src-d/code-annotation/pull/10#issuecomment-359037486 - it seems that current status of this is WIP due to

> The requirements for that connection URL are basically the ability to connect to SQLite or PostgreSQL.

carlosms:
Code in master. Any future improvements will be done in a separate issue.

PR #10, opened 2018-01-15
First version of the import tool

carlosms:
This is an initial version of the import tool, described in issue #4.
The main functionality is implemented, but it needs work.

A couple of comments on the choices made:
- To access sqlite DB I've chosen the library github.com/mattn/go-sqlite3, as it seems mature and the de-facto option.
- There might be a better way to perform the diff, but at this stage I didn't want to test a bunch of different go libraries and just went with the diff executable and tmp files.


tools/import.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func printHelp() {
+	// TODO: Improve help message
+	fmt.Println("Usage: import <path-to-origin.db> <path-to-destination.db>")
+}
+
+// bootstrap creates the necessary tables for the output DB. It is safe to call on a
+// DB that is already bootstrapped.
+func bootstrap(db *sql.DB) error {
```
smacker: I don't have the strong opinion here. SQL in code is fine for me, but as an option in gemini we have sql text file with a schema. And execute it when needed.

tools/import.go
```diff
@@ -0,0 +1,207 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// importFiles imports pairs of files from the origin to the destination DB.
+// It copies the contents and processes the needed data (md5 hash, diff)
+func importFiles(originDB, destDB *sql.DB) (nFiles int64, err error) {
+	// TODO: consider using transaction for increased speed
```
smacker: yeah. sqlite commit after each statement by default. It would be nice to commit in batches. But maybe not right now. Up to you.

tools/import.go
```diff
@@ -0,0 +1,207 @@
...
+		var nameA, nameB, contentA, contentB string
+		if err := rows.Scan(&nameA, &nameB, &contentA, &contentB); err != nil {
+			log.Fatal(err)
+		}
+
+		res, err := insert.Exec(nameA, nameB,
+			md5hash(contentA), md5hash(contentB),
+			contentA, contentB,
+			diff(contentA, contentB),
+			2) // TODO: set a valid experiment_id
```
smacker: can we agree to hardcode id 1?

tools/import.go
```diff
@@ -0,0 +1,207 @@
...
+
+	return
+}
+
+func md5hash(text string) string {
+	return fmt.Sprintf("%x", md5.Sum([]byte(text)))
+}
+
+// tmpWrite creates a temporary file and writes the given text in it
+func tmpWrite(text string) (*os.File, error) {
```
smacker: I really think it isn't a good idea to use tmp files for diff. `git-go` uses https://github.com/sergi/go-diff Api looks simple. If it does the job I would just use it.

tools/import.go
```diff
@@ -0,0 +1,207 @@
+/*
```
smacker: in Go you usually put executables in cmd/_name_/_name_.go It's kind of convention.

smacker (changes requested):
few comments.

tools/import.go
carlosms: done

tools/import.go
carlosms: For now I've left it in the code since it's a small amount of sql statements, but I'm fine changing it if others think that's better

tools/import.go
carlosms: done

tools/import.go
carlosms: done

tools/import.go
carlosms: That lib appears to not implement unified diff output, I've used this one instead: https://github.com/pmezard/go-difflib

smacker (approved):

cmd/import/import.go
```diff
@@ -0,0 +1,223 @@
...
+
+func main() {
+	args := os.Args
+
+	if len(args) != 3 {
+		printHelp()
+		return
+	}
+
+	originPath := args[1]
```
dpordomingo: It would be great if we use [go-flags](https://github.com/jessevdk/go-flags) as in most of our commands.
[Example in borges](https://github.com/src-d/borges/blob/master/cli/borges/main.go#L93)

cmd/import/import.go
```diff
@@ -0,0 +1,223 @@
...
+
+Both arguments refer to SQLite databases. If the destination file does not
+exist, it will be created.
+The destination database does not need to be empty, new imported file pairs can
+be added to previous imports.`)
+}
+
+// bootstrap creates the necessary tables for the output DB. It is safe to call on a
+// DB that is already bootstrapped.
+func bootstrap(db *sql.DB) error {
```
dpordomingo: I'd move the DB creation outside of the command package

cmd/import/import.go
```diff
@@ -0,0 +1,223 @@
...
+	_, err := db.Exec(`
+		CREATE TABLE IF NOT EXISTS users (
```
dpordomingo: If you define the SQL commands as constants, then you could do, more concisely:
```go
	if _, err := db.Exec(createUsers); err != nil {
		return err
	}
    ...
	return nil

```

cmd/import/import.go
```diff
@@ -0,0 +1,223 @@
...
+
+func diff(nameA, nameB, contentA, contentB string) (string, error) {
+	diff := difflib.UnifiedDiff{
+		A:        difflib.SplitLines(contentA),
+		B:        difflib.SplitLines(contentB),
+		FromFile: nameA,
+		ToFile:   nameB,
+		Context:  3,
+	}
+	text, err := difflib.GetUnifiedDiffString(diff)
```
dpordomingo: you can delete everything below this line if it'd be replaced by 
```go
	return difflib.GetUnifiedDiffString(diff)
```

cmd/import/sample.sql
```diff
@@ -0,0 +1,755 @@
+/*
+Sample DB for the import command. You can create an sqlite DB with the command:
```
dpordomingo: I'd move this to `cmd/examples/import/example.sql`

cmd/import/import.go
```diff
@@ -0,0 +1,223 @@
...
+	_, err = db.Exec(`
+		CREATE TABLE IF NOT EXISTS experiments (
+			id INTEGER, name TEXT UNIQUE, description TEXT,
+			PRIMARY KEY (id))`)
+
+	if err != nil {
+		return err
+	}
+
+	// TODO: consider a unique constrain to avoid importing identical pairs
```
dpordomingo: Why not adding it?
```UNIQUE(hash_a, hash_b)```

**Review by dpordomingo — CHANGES_REQUESTED** [review]
I find it globally quite good, so congrats !!!

I'd consider some things before merging:
- using [go-flags](https://github.com/jessevdk/go-flags) as in most of our commands,
- moving the DB handling into a separate package, so the command will take care only of the cli thing,
- moving the examples to a examples directory.

File: cmd/import/import.go [hunk 12]
```diff
@@ -0,0 +1,223 @@
+/*
```
dpordomingo: For internal consistency &ndash;as discussed via Slack&ndash; I'd rename `cmd` by `cli`

dpordomingo (changes requested):
renaming `cmd` dir

dpordomingo (changes requested):
I didn't realize. I miss the tests. ;)
(most of our projects are using [testify](http://github.com/stretchr/testify))

smacker:
I'm not sure we need tests for it. We don't know schema of input DB. And there is almost no code beside transforming one schema to another.

dpordomingo:
As agreed _irl_, the tests can be quite simple and only for the internal part of the command.
It will be useful to @carlosms to know more about them in go, and it will give us a starting point when we have a new schema.

cmd/import/sample.sql
```diff
@@ -0,0 +1,755 @@
...
+
+$ sqlite3 sample.db < sample.sql
+*/
+CREATE TABLE files (name_a TEXT, name_b TEXT, content_a TEXT, content_b TEXT);
```
bzz: Most probably, eventually we will want to keep more information on files i.e
 - `repository_id` they are coming from
 - `commit_hash` inside that repository
 - path
 - `blob_id` sha1 hash of the blob

But this can be taken care of in subsequent PRs.

bzz:
👍 @carlosms great job.

One thing though - WIP in title of PR usually means `Work in Progress - stay out, it's not ready for reviews` so may be that way it's not that WIP 😉 

On reviews:
 - I agree with @smacker. At the same time, it would be soo nice to have at least one very simple test example, just as an infrastructural foundation for future work.

 - Using go-flags is a nice idea for the same reason and is quite low-hanging fruit. 

 - Extracting constants for string literals is a good engineering practice as well.

For me, extra work of moving things in FS (including DB handling) could be done in subsequent PRs, together with more useful work.

It's up to you, I would say it's ready to be merged and improved in subsequent PRs as soon as those 3 things are somehow improved.

One last thing is - I would suggest everyone to setup commit sign-off \w DCO to comply with [src-d contributing guideline](https://github.com/src-d/guide/blob/master/engineering/documents/CONTRIBUTING.md#certificate-of-origin). 

After doing this, something like `git rebase --exec 'git commit --amend --no-edit -n -S' -i HEAD~<N>` would sign last N commits for you.

While at it, it may be worth looking into [commit signing](https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work) or `-s` which is not mandatory, but easy to add as well.

cmd/import/sample.sql
dpordomingo: Yes, we can easily add that properties or they can even use the name for that (it would not require changes in our app), for example:
they can name the files as: `<repo>/blob/<commit>/<path>`, so it would include the repository, commit, and path, example:
`github.com/src-d/go-git/blob/110e7012e498c33ba2cbc866f60bb52edca947ae/README.md`
.

cmd/import/import.go
carlosms: done

cmd/import/import.go
carlosms: done

cmd/import/import.go
carlosms: done

cmd/import/sample.sql
carlosms: done

cmd/import/import.go
carlosms: done in https://github.com/src-d/code-annotation/pull/10/commits/20e8e3f13e90268221a1b05f4cf1fd42cbc4d4b2

bzz:
@carlosms nice! 

Feel free to post a comment here, describing the status of the feedback that was provided as soon as you see fit, and I'll be happy to make another pass on it.

cmd/import/import.go
carlosms: done in https://github.com/src-d/code-annotation/pull/10/commits/aee4d8ea3f254b4947dce8b5153a8c063b61463e & https://github.com/src-d/code-annotation/pull/10/commits/31941005ef472b4fa94a5eb260c707088d74c4b6

cmd/import/import.go
carlosms: I would wait until we have real input examples. Maybe two pairs with the names and the same hashes point to identical files in different repos.

server/dbutil/dbutil.go
```diff
@@ -0,0 +1,161 @@
...
+		createAssignments = `CREATE TABLE IF NOT EXISTS assignments (
+			user_id INTEGER, pair_id INTEGER, experiment_id INTEGER,
+			answer INTEGER, duration INTEGER,
+			PRIMARY KEY (user_id, pair_id),
+			FOREIGN KEY (user_id) REFERENCES users(id),
+			FOREIGN KEY (pair_id) REFERENCES file_pairs(id),
+			FOREIGN KEY (experiment_id) REFERENCES experiments(id))`
+	)
+
+	if _, err := db.Exec(createUsers); err != nil {
```
smacker: I would use a loop here. But up to you.

cmd/import/sample.sql
carlosms: I agree, but until we have real input examples, this file assumes the bare minimum data to start developing.

carlosms:
@bzz @dpordomingo Most of the feedback has been addressed now.
The only pending request is the creation of tests.

server/dbutil/dbutil.go
carlosms: What would be the idiomatic way in Go? How about this:

	tables := []string{createUsers, createExperiments, createFilePairs, createAssignments}

	for _, table := range tables {
		if _, err := db.Exec(table); err != nil {
			return err
		}
	}

smacker: yes. I would write something like this to avoid multiple copy-past of `db.Exec; if err`.
There are many other ways how to write it, but I would do it as simple as possible. (your snippet above)

server/dbutil/dbutil.go
```diff
@@ -0,0 +1,161 @@
...
+			logger.Println(err)
+			failures++
+			continue
+		}
+
+		rowsAffected, _ := res.RowsAffected()
+		success += rowsAffected
+	}
+
+	if err := tx.Commit(); err != nil {
```
smacker, line 145: do you want to add `rollback` somewhere? I'm not sure about sqlite, but for other dbs it's better to rollback. :)

carlosms: It actually looks like Commit or Rollback should be the last call, if I understand this correctly Commit will close the transaction even if it fails:
https://golang.org/src/database/sql/sql.go?s=49331:49359#L1761

carlosms:
Basic tests pushed. I'd say there are no outstanding issues.

server/dbutil/dbutil.go
smacker: You do `return success, failures, err` above. So neither Commit neither Rollback will be executed. And we will end up with hanging transaction.

carlosms: Yes, missed that return inside the loop, thank you.

server/dbutil/dbutil.go
```diff
@@ -0,0 +1,156 @@
...
+		return 0, success + failures, err
+	}
+
+	if err := rows.Err(); err != nil {
+		return success, failures, err
+	}
+
+	return success, failures, nil
```
bzz: probably a nitpick, but any reason why not

```go
return success, failures, rows.Err()
```
?

server/dbutil/dbutil.go
```diff
@@ -0,0 +1,156 @@
...
+	_ "github.com/mattn/go-sqlite3"
+	"github.com/pmezard/go-difflib/difflib"
+)
+
+const defaultExperimentID = 1
+
+// Bootstrap creates the necessary tables for the output DB. It is safe to call on a
+// DB that is already bootstrapped.
+func Bootstrap(db *sql.DB) error {
+	const (
```
bzz: Any reason why not to extract all query strings from everywhere to the top-level constants all together in a single place?

carlosms: Personal preference, I guess. I like to see the query contents near to where it is been used. But if the common style in other tools is to have all SQL queries as const at the top, that is also fine for me. Should I go ahead and move all queries?

server/dbutil/dbutil.go
carlosms: done

cli/import/import.go
```diff
@@ -0,0 +1,86 @@
...
+
+	"github.com/jessevdk/go-flags"
+	"github.com/src-d/code-annotation/server/dbutil"
+)
+
+const defaultExperimentID = 1
+
+const desc = `Imports pairs of files from the input database to the output database.
+If the destination file does not exist, it will be created.
+The destination database does not need to be empty, new imported file pairs can
```
dpordomingo: reading 
_>"new imported file pairs can be added to previous imports"_
I understood that running the import command will import only "new pairs" &rarr; not already imported,
but it is not true because of the missing `UNIQUE` constraint.

I'd propose one of the following:
- rewording saying that import command will add all source pairs regardless they were already imported,
- adding a `UNIQUE` constrain to ensure that only new pairs will be added.

cli/import/import.go
```diff
@@ -0,0 +1,86 @@
...
+package main
+
+import (
+	"database/sql"
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/jessevdk/go-flags"
+	"github.com/src-d/code-annotation/server/dbutil"
```
dpordomingo: external dependencies and internal dependencies should be separated with one line:
```go
import (
    // stdlib ones

    // internal dependencies

    // external dependencies
)
```
example in [go-git](https://github.com/src-d/go-git/blob/master/common_test.go)

server/dbutil/dbutil.go
dpordomingo: I'm used to see them at package level too

server/dbutil/dbutil.go
```diff
@@ -0,0 +1,152 @@
...
+
+// Initialize populates the DB with default values. It is safe to call on a
+// DB that is already initialized
+func Initialize(db *sql.DB) error {
+	_, err := db.Exec(`
+		INSERT OR IGNORE INTO experiments (id, name, description)
+		VALUES ($1, 'default', 'Default experiment')`,
+		defaultExperimentID)
+
+	if err != nil {
```
dpordomingo: `return err` would be enough, no `if` needed

server/dbutil/dbutil.go
```diff
@@ -0,0 +1,152 @@
...
+// Options for the ImportFiles method. Logger is optional, if it is not provided
+// the default stderr will be used.
+type Options struct {
+	Logger *log.Logger
+}
+
+// ImportFiles imports pairs of files from the origin to the destination DB.
+// It copies the contents and processes the needed data (md5 hash, diff)
+func ImportFiles(originDB, destDB *sql.DB, opts Options) (success, failures int64, e error) {
+	logger := log.New(os.Stderr, "", log.LstdFlags) // Default log to stderr
```
dpordomingo: The logger can be overrided in the next `if`
What if it's done explicitely:
```go
var logger *log.Logger;
if opts.Logger != nil {
    logger = opts.Logger
} else {
    logger = log.New(os.Stderr, "", log.LstdFlags)
}
```

server/dbutil/dbutil.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	if err := tx.Commit(); err != nil {
+		return 0, success + failures, err
+	}
+
+	return success, failures, rows.Err()
+}
+
+func md5hash(text string) string {
```
dpordomingo, line 152: I'd name it `hash` or something similar, so its implementation would not depend on the function name.

server/dbutil/dbutil_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	for _, hash := range hashes {
+		assert.Equal(suite.T(), hash[1], md5hash(hash[0]))
+	}
+}
+
+func readFile(filename string) string {
+	data, err := ioutil.ReadFile(filename)
+	if err != nil {
+		panic(err)
```
dpordomingo: `panic` is not needed here. The `readFile` func could return `(string, error)`, and the error would be validated with `assert.NoError(err)`. That way the tests will break gracefully.

dpordomingo (changes requested):
Wow! I like it a lot.
I have only one major requirement ([the `panic` one](https://github.com/src-d/code-annotation/pull/10#discussion_r162421446)) and a couple of minor suggestions.

cli/import/import.go
carlosms: fixed

cli/import/import.go
carlosms: I added a note in the command description. I still think adding a unique clause to a hypothetical input schema is a premature optimization, we can change this later.

server/dbutil/dbutil.go
carlosms: fixed

server/dbutil/dbutil.go
carlosms: changed. I initialized it with the default value assuming that the option was not going to be used commonly.

server/dbutil/dbutil.go
carlosms: Well, in this case the function is not really returning _any_ hash, it is computing the MD5 hash. A change in the algorithm is not really a local change, as it may break other functionality in the app.
I'd rather leave like it is, if you agree.

server/dbutil/dbutil_test.go
carlosms: I'm going to make the change you propose, because I see your point, but at the same time those files are part of the repo and the test.
If the file is not there because I deleted it accidentally, doesn't it make sense to return a severe failure (`panic`) instead of a failed test?

cli/import/import.go
dpordomingo: option a) great

server/dbutil/dbutil.go
dpordomingo: fair enough :+1:

server/dbutil/dbutil_test.go
dpordomingo: Panics are always the last last last option.
In my opinion, for the purpose of a test, it's enough with failing and keep running the other tests.

dpordomingo (approved):
Great!!!
LGTM

don't forget the squash/rebase/sign all commits before merging ;)

bzz (approved):
LGTM.

Merge at will, either now or after testing/adjusting schema on real data.

bzz (changes requested):
Sorry, just realized that at least one input argument in this one needs to be changed and aligned with https://github.com/src-d/code-annotation/pull/15#issuecomment-358938344

Basically, the idea is to keep application DB abstracted of RDBMS implementation used by passing around full connection URL in agnostic format like dialect+driver://username:password@host:port/database

carlosms:
As discussed IRL, this PR will be merged and the functionality to provide a DB connection URL will be done in a future PR.
The requirements for that connection URL are basically the ability to connect to SQLite or PostgreSQL.

PR #14, opened 2018-01-18
WIP: Change docker image name

vmarkovtsev:
As Francesc suggested in https://github.com/src-d/devrel/issues/1

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@ In order to evaluate quality of ML models, as well as to create “ImageNet for 
 ### Docker
 
 ```bash
-docker build -t code-annotation .
-docker run --rm -p 8080:8080 code-annotation
+docker build -t src-d/code-annotation .
+docker run --rm -p 8080:8080 src-d/code-annotation
 ```
 
 ### Non-docker
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@ In order to evaluate quality of ML models, as well as to create “ImageNet for 
 ### Docker
 
 ```bash
-docker build -t src-d/code-annotation .
-docker run --rm -p 8080:8080 src-d/code-annotation
+docker build -t srcd/code-annotation .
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,8 +7,8 @@ In order to evaluate quality of ML models, as well as to create “ImageNet for
 ### Docker
 
 ```bash
-docker build -t code-annotation .
-docker run --rm -p 8080:8080 code-annotation
+docker build -t src-d/code-annotation .
```
dpordomingo: But this is the name of the image when locally created.
Is this really required by https://github.com/src-d/devrel/issues/1 ?

vmarkovtsev: Sure. I mean that our docker name is src-d and we suppose that there is no difference between a local image and the dockerhub's.

dpordomingo: fair enough

dpordomingo (approved):
lgtm

bzz:
@vmarkovtsev thank you for contribution! Wonder, if that is that a cross-organizational effort, like are we changing it in all repos? Would prefer to do that all together.

Also https://hub.docker.com/u/src-d/ results in 404 to me, so I suggest we rename it only after the whole thing works. Have added WIP in the PR title, until dockerHub org is renamed.

But, please, let me know if I'm missing something here.

vmarkovtsev:
This is just my general notice how the containers have been named de-facto, albeit srcd instead of src-d. The rest is up to @campoy @marnovo

bzz:
Got it, thank you for explanation!
If you would not mind changing `src-d -> srcd` for now until dockerHub org renamed, we could merge this asap.

vmarkovtsev:
IMHO it is fine to name it srcd right now because Francesc or Marcelo will rename the whole org soon.

bzz (approved):

Issue #5, opened 2018-01-12
Tool for exporting annotations


bzz:
A tool that can be called from CLI and an HTTP handler, that produces sqlite DB with pairs of files + user + answer (any schema for now, will be changed later) from an internal DB defined in #2 .

Output schema should be as simple as possible and allow to query for all answers of the single author, as well as to get all files with it's annotations.

As soon as there is something to share, we could show it to @EgorBu and iterate on his feedback (updating Design Doc).

carlosms:
Issue blocked until we have the required output DB schema

bzz:
Let's move forward with this and just suggest a simples schema that allows:
 - to query for all answers of the single author
 - get files with it's annotations

and iterate based on actual user feedback from there.

carlosms:
> Let's move forward with this and just suggest a simples schema that allows:
> 
> - to query for all answers of the single author
> - get files with it's annotations

Then I think the current version works fine, a copy of the internal DB already provides those two features.

bzz:
@carlosms sounds good, so then this is still open due to work done to support SQLight/Postgres connection URL as source argument in exporter?

carlosms:
That's right, I just created PR #28 with code for the import tool.
The changes for the export tool will be pushed to that PR also.

carlosms:
Code in master. Any future improvements will be done in a separate issue.

PR #15, opened 2018-01-18
Placeholder for the export tool

carlosms:
Initial work on #5.

This is a placeholder that will be finished when the output schema is defined. For now, the tool does a simple copy&paste of the internal DB SQLite file.

This PR does not have any dependencies (as opposed to what I said initially IRL).

### DIFF
--- a/cli/export/export.go
+++ b/cli/export/export.go
@@ -62,7 +62,7 @@ func copy(source, destination string) error {
 	}
 	defer src.Close()
 
-	dst, err := os.OpenFile(destination, os.O_RDWR|os.O_CREATE|os.O_TRUNC|os.O_EXCL, 0666)
+	dst, err := os.OpenFile(destination, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0666)
 	if err != nil {
 		return err
 	}
--- a/cli/export/export.go
+++ b/cli/export/export.go
@@ -1,0 +1,76 @@
+/*
+Tool to export annotation results from the internal DB to an output DB.
+
+Usage: export <path-to-origin.db> <path-to-destination.db>
+*/
+package main
+
+import (
+	"fmt"
+	"io"
+	"log"
+	"os"
+
+	"github.com/jessevdk/go-flags"
+)
+
+const desc = `Exports annotation results from the internal input database to a new output
+database.
+The destination file must not exist.`
+
+var opts struct {
+	Args struct {
+		Input  string `description:"SQLite database filepath"`
+		Output string `description:"SQLite database filepath"`
+	} `positional-args:"yes" required:"yes"`
+}
+
+func main() {
+	parser := flags.NewParser(&opts, flags.Default)
+	parser.LongDescription = desc
+
+	if _, err := parser.Parse(); err != nil {
+		if err, ok := err.(*flags.Error); ok {
+			if err.Type == flags.ErrHelp {
+				os.Exit(0)
+			}
+
+			fmt.Println()
+			parser.WriteHelp(os.Stdout)
+		}
+
+		os.Exit(1)
+	}
+
+	if _, err := os.Stat(opts.Args.Input); os.IsNotExist(err) {
+		log.Fatalf("File %q does not exist", opts.Args.Input)
+	}
+
+	// TODO: until we have more information about the output schema needed,
+	// this tool does a simple copy&paste of the internal DB. The annotation
+	// results are stored in the assignments table.
+
+	if err := copy(opts.Args.Input, opts.Args.Output); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func copy(source, destination string) error {
+	src, err := os.Open(source)
+	if err != nil {
+		return err
+	}
+	defer src.Close()
+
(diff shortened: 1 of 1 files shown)


cli/export/export.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func copy(source, destination string) error {
+	src, err := os.Open(source)
+	if err != nil {
+		return err
+	}
+	defer src.Close()
+
+	dst, err := os.OpenFile(destination, os.O_RDWR|os.O_CREATE|os.O_TRUNC|os.O_EXCL, 0666)
```
dpordomingo: if `os.O_CREATE|os.O_EXCL` requires that the file does not exist,
then `os.O_TRUNC` is not necessary, is it?

cli/export/export.go
```diff
@@ -0,0 +1,76 @@
...
+			}
+
+			fmt.Println()
+			parser.WriteHelp(os.Stdout)
+		}
+
+		os.Exit(1)
+	}
+
+	if _, err := os.Stat(opts.Args.Input); os.IsNotExist(err) {
```
dpordomingo, line 45: I think this is not neccesary, because when in [`if copy() != nil`](https://github.com/src-d/code-annotation/pull/15/files#diff-fe387ed169ad94895f58f2465e5359cdR53) it's already validated that there is no errors because the source can not be [opened](https://github.com/src-d/code-annotation/pull/15/files#diff-fe387ed169ad94895f58f2465e5359cdR59) or [readed](https://github.com/src-d/code-annotation/pull/15/files#diff-fe387ed169ad94895f58f2465e5359cdR71)

cli/export/export.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	defer src.Close()
+
+	dst, err := os.OpenFile(destination, os.O_RDWR|os.O_CREATE|os.O_TRUNC|os.O_EXCL, 0666)
+	if err != nil {
+		return err
+	}
+	defer dst.Close()
+
+	if _, err := io.Copy(dst, src); err != nil {
```
dpordomingo, line 71: The `if` is not necesary, you can also do:
```go
_, err := io.Copy(dst, src)
return err
```

dpordomingo (commented):
Quite good, thanks!
I have only one question, and a couple of suggestions :D

cli/export/export.go
carlosms: True. But this is here in preparation for when the command uses this path to open an SQLite connection. This check will avoid creating a new file.

dpordomingo: fair enough

cli/export/export.go
carlosms: You are right, I just copied the call made in os.Create and added O_EXCL without much care... fixed now.

cli/export/export.go
carlosms: Thanks. To be honest, somehow my version feels cleaner to me. Separating the error handling and the last return with no error makes it easier to drop new lines by someone later. Otherwise the new author may forget the `if err!=nil return err`; and the new diff will show modifications to io.Copy when that is not really part of the new functionality.

Having said that, I'm not against this change. Is this considered the idiomatic way, or is it more a team convention?

smacker: the standard library is the best place for inspiration:
https://github.com/golang/go/blob/master/src/net/http/response.go#L321
https://github.com/golang/go/blob/master/src/time/time.go#L1232

both versions are valid. I'm ok with extra checking for an error and returning nil otherwise. But I know some people don't like it.

smacker (approved):

cli/export/export.go
dpordomingo: Removing the `if` doesn't mean that the error is not handled (I feel it's even more explicit that we're returning the error returned by the `io.Copy`). Once said that, I don't see the benefit of a redundant `if` when it's not necessary (considering that Go design does not focus in keeping the diffs as smaller as possible after every code modification)

Anyhow, since the stdlib is not consistent as Smaker pointed out, feel free to proceed on your own ;)

dpordomingo (approved):
Considering https://github.com/src-d/code-annotation/pull/15#discussion_r162536558 you can proceed as you consider the better way.

In any case, you can merge after squash :)

bzz:
@carlosms good start!

I think it would be even more useful to have this tool, instead of relaying on internal implementation details of how DB is managed in this web app (SQLight), abstract the source of data through a well-define interface of standard URL?

How about changing it, so it has as input a URL for web-app DB to connect to (any RDBMS) like `dialect+driver://username:password@host:port/database` and an output an SQLight file with some simple schema.

So in this case, input can be like `sqlite:////absolute/path/to/foo.db`.

bzz (changes requested):
`path-to-origin.db` needs to be replaced by URL

cli/export/export.go
smacker: I'm ok to always remove redundant `if`s if everybody else wants it. Should we vote?

carlosms:
Same comment as https://github.com/src-d/code-annotation/pull/10#issuecomment-359037486

As discussed IRL, this PR will be merged and the functionality to provide a DB connection URL will be done in a future PR.
The requirements for that connection URL are basically the ability to connect to SQLite or PostgreSQL.

Issue #2, opened 2018-01-12
Schema design


bzz:
Define SQL schema that is required to store 
 - files
 - questions
 - answers

dpordomingo:
That's what @carlosms and I would propose
```
- Users # They're the actors of the platform
  - ID
  - github_username
  - auth
  - role # [Requesters|Workers]

- FilePairs # They're the questions of the experiments, where user compares two source codes
  - ID
  - name_A # path
  - name_B # path
  - hash_A # MD5, SHA1... of the content
  - hash_B # MD5, SHA1... of the content
  - content_A
  - content_B
  - experiment_ID
  - diff

- Experiments # They describe the experiments where user compares two source codes.
  - ID
  - name
  - description

- Assignment # It stores the user answers
  - user_ID
  - pair_ID
  - experiment_ID
  - answer # [Yes|KindOf|Nope|Skip|Null]
  - duration # seconds
```

**Assumptions:**
- All Assignments are created when the Worker starts answering an Experiment, being `Null` the value of the answer, and `0` the time elapsed,
- If a worker changes its mind, and answer again an Assignment, the answer will be updated, and the new duration will be summed to the old one (or updated too, if requirements changes)

**Simplifications:**
- We tried to model the DB considering only what we want to build, avoiding more complex possibilities like making the `FilePairs` a subtype of `Question`, having same `FilePairs` in different `Experiments` so we'd need a crossing `ExperimentFilePairs`, and so on.

dpordomingo:
@smacker already validated it irl; @bzz feel free to reopen if it's needed to change something

carlosms:
I'm reopening this issue to inform you of a small change that I've made after getting real input data. Please take a look, specially @dpordomingo since you are working on the backend code, and then we can close this issue if you agree.

The import tool merged to master already creates the DB with this schema change.

Now the `file_pairs` table has the following columns:

- id
- blob_id_a
- repository_id_a
- commit_hash_a
- path_a
- content_a
- hash_a
- blob_id_b
- repository_id_b
- commit_hash_b
- path_b
- content_b
- hash_b
- score
- diff
- experiment_id

dpordomingo:
Great. Thanks @carlosms
I'll add it to https://github.com/src-d/code-annotation/pull/16
Feel free to reopen if there's any news

PR #16, opened 2018-01-18
Backend Api mocks

### LINKED ISSUE #3: Backend: REST API

dpordomingo:
partially solves #3

This PR will create a backend api serving mocks, using the requirements defined by https://github.com/src-d/code-annotation/issues/3

If approved, the things that will be done in the next PRs will be:
- using DB stuff,
- tests,
- improve error handling and error messages,
- using user from Auth


server/cmd/code-annotation/serializer/serializers.go
```diff
@@ -0,0 +1,112 @@
...
+
+type ResponseMany struct {
+	Data []Identifable `json:"data"`
+}
+
+func (e *ResponseMany) Render(w http.ResponseWriter, r *http.Request) error {
+	return nil
+}
+
+type Identifable interface {
```
carlosms: typo? identifiable

carlosms (approved):

server/cmd/code-annotation/serializer/serializers.go
dpordomingo: ouch! many thanks

bzz:
Nice!

A bit confused though - what is the reason to have all the code under `/cmd/` package? 

Usually, it is reserved for executable things, like server of CLI, etc. (see [borges/cli](https://github.com/src-d/borges/tree/master/cli/borges)). Everything else could belong to a better-named package under `/server/`.

smacker:
Could you please not merge it yet. I want to take a look. I'll do it a bit later today. Thanks.

server/cmd/code-annotation/handler/assignments.go
```diff
@@ -0,0 +1,74 @@
...
+	"encoding/json"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"strconv"
+
+	"github.com/go-chi/chi"
+	"github.com/go-chi/render"
+
+	"github.com/src-d/code-annotation/server/cmd/code-annotation/serializer"
```
smacker: do you think it's better to move it out of cmd? It's a common pattern in go to keep only entry point in cmd folder.
Please let me know if you need more information about it.

server/cmd/code-annotation/handler/assignments.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/src-d/code-annotation/server/model"
+	"github.com/src-d/code-annotation/server/repository"
+)
+
+func HandleGetAssignmentsForUserExperiment(w http.ResponseWriter, r *http.Request) {
```
smacker: do you think we could follow https://github.com/golang/go/wiki/CodeReviewComments#package-names here?
The package has name `handler`. So I would suggest `handler.GetAssignmentsByUser`. I think we don't really need `experiment` in the name because looks like assignments always belong to experiment. Feel free to suggest better name!

server/cmd/code-annotation/handler/assignments.go
```diff
@@ -0,0 +1,74 @@
...
+	experimentID, err := strconv.Atoi(chi.URLParam(r, "experimentId"))
+	if err != nil {
+		render.Render(w, r, serializer.NewErrors(err))
```
smacker: do you think we can provide better error message here? Currently, it's `strconv.Atoi: parsing "qwe": invalid syntax`. I would suggest add information that we can't parse `experimentId` not a random string.

server/cmd/code-annotation/handler/assignments.go
```diff
@@ -0,0 +1,74 @@
...
+		return
+	}
+
+	userID, err := strconv.Atoi("1")
+	if err != nil {
+		render.Render(w, r, serializer.NewErrors(err))
+		return
+	}
+
+	assignments, err := repository.GetAssignmentsFor(userID, experimentID)
```
smacker: As I remember here we create all assignments. How do you think is it better to put creation logic here or in a service? And keep repository as a slim interface to DB?

server/cmd/code-annotation/handler/assignments.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	body, err := ioutil.ReadAll(r.Body)
+	defer r.Body.Close()
+	if err != nil {
+		render.Render(w, r, serializer.NewErrors(err))
+		return
+	}
+
+	var assignment serializer.Assignment
```
smacker: I usually try to keep structures close to the place where they are used (in the same package and as a private struct) if they are used only once. What is your opinion about it?

server/cmd/code-annotation/handler/assignments.go
```diff
@@ -0,0 +1,74 @@
...
+	if err := json.Unmarshal(body, &assignment); err != nil {
+		render.Render(w, r, serializer.NewErrors(err))
+		return
+	}
+
+	answer := model.Answer(assignment.Answer)
+	if err := repository.SaveAssignment(assignmentId, answer, assignment.Duration); err != nil {
```
smacker: Usually repository save methods accept model. But up to you.

server/cmd/code-annotation/handler/assignments.go
```diff
@@ -0,0 +1,74 @@
...
+		render.Render(w, r, serializer.NewErrors(err))
+		return
+	}
+
+	//TODO: use chi logger
```
smacker: can we use logrus?

dpordomingo:
Sure!
I think it's a good idea to avoid merging things before they're fully reviewed by all of us.

server/cmd/code-annotation/serializer/serializers.go
```diff
@@ -0,0 +1,112 @@
+package serializer
+
+import (
+	"net/http"
+
+	"github.com/go-chi/render"
```
smacker: render package is quite complex. The main advantage of it, you can handle different types of input/output. Like json/xml for example. But I'm not sure we really need it in this project.
Are you open to considering simpler solution?

server/cmd/code-annotation/serializer/serializers.go
```diff
@@ -0,0 +1,112 @@
...
+type ResponseStored struct {
+	Data int `json:"data"`
+}
+
+func (e *ResponseStored) Render(w http.ResponseWriter, r *http.Request) error {
+	return nil
+}
+
+type ResponseMany struct {
+	Data []Identifiable `json:"data"`
```
smacker: do you think it's possible we will render something that doesn't have id?
I personally prefer to add constraints only when they are really necessary.
please consider this comment only as an open discussion, not action request.

server/model/models.go
```diff
@@ -0,0 +1,54 @@
+package model
+
+type User struct {
```
smacker, line 3: we defined this model in schema task, but it doesn't have all necessary field. For example avatar url.
I would also like to abstract from github in model. `login` and `username` fields are better in my opinion. Then we can add other providers later. Google for example. Or allow registering user. If you use abstract model consumers of model don't need to know how user was created.
Please share your thought!

server/model/models.go
```diff
@@ -0,0 +1,54 @@
...
+
+const (
+	Requester Role = "requester"
+	Worker    Role = "worker"
+)
+
+type Answer string
+
+const (
+	Yes   Answer = "yes"
```
smacker: On one hand `yes, no, maybe` sounds common enough. On another hand, I'm not 100% sure it's the best options for our tool.
I remember you spent time trying other tools. Could you please remind me what do they use as names for action?
It's very basic part of the tool, I just want to do it right. Thanks!

smacker (changes requested):
Good job. But I have comments. Some of them are actually applicable to many similar lines, but I didn't copy-past them to avoid bloating of PR.

dpordomingo:
answering @bzz question https://github.com/src-d/code-annotation/pull/16#issuecomment-358920441
I tried to follow the [initial skeleton](https://github.com/src-d/code-annotation/pull/7) but you're right, I'll ping you when it had been reordered.
It would be:
```bash
/
  public # web stuff
  src # web sources
  cli
    import/import.go
    export/export.go
    server/server.go
    examples/import/example.sql # it may be moved to of /cli/import/examples/example.sql
  server
    handler
    serializer
    model
    repository
    dbutil # it may be moved outside of /server (to the root path)
```

Comment with :+1: if you agree, or provide a new suggestion :wink:

server/cmd/code-annotation/handler/assignments.go
dpordomingo: Sure!!!
That aligns quite well with [Alex comment](https://github.com/src-d/code-annotation/pull/16#issuecomment-358920441)
I tried to follow the initial skeleton and I might misunderstand it.
I'd propose the following https://github.com/src-d/code-annotation/pull/16#issuecomment-358947367

server/cmd/code-annotation/handler/assignments.go
dpordomingo: I'd sugest `handler.HandleInHandlerPackageThatGetAssignmentsByUserExperiment`
:trollface: 

You're totally right.

I only disagree in the `Experiment` suffix, because there will be many experiments; that's because handled URL is `/experiments/{experimentId}/assignments`.

server/cmd/code-annotation/handler/assignments.go
dpordomingo: sure thing!
That's because I left as "pending" in the PR desc:
>_improve error handling and error messages_

I just tried to postpone the error messages for the next iteration, so it wouldn't block the API.
If you agree, we'd merge without this "feature".

server/cmd/code-annotation/handler/assignments.go
dpordomingo: I was thinking in:
```go
assignments, err := repository.GetAssignmentsFor(userID, experimentID)
if err.is(NO_ASSIGNMENTS_INITIALIZED) {
    assignments, err = repository.CreateAssignmentsFor(userID, experimentID)
} 
```
I agree on putting business logic in `service`, instead of in `repository`... but should we create a new `service` package for only these 3 lines that could be in the `handler.GetAssignmentsForUserExperiment` ?

server/cmd/code-annotation/handler/assignments.go
dpordomingo: I think that handlers and serializers should be separated.

server/cmd/code-annotation/handler/assignments.go
dpordomingo: Would you like something like


```go
package repository

func UpdateAssignment(id int, answer model.Assignment) error
```
and then use it as below?
```go
if err := repository.UpdateAssignment(assignmentId, assignmentPayload); err != nil {
```

&ast; I send the `assignmentId`, because I'd not rely in the `assignmentPayload`

server/cmd/code-annotation/handler/assignments.go
dpordomingo: I'd use chi default through `chi/middleware/Logger`, whose default seems to be `STDOUT` and `STDERR`, but also lets us use Logrus or others.

I'd prefer to consider the Logging lib to use in a separate issue since we're using three big logging libraries in different of our biggest projects
- `log` &rarr; go-git...
- `github.com/inconshreveable/log15` &rarr; borges, rovers, framework...
- `github.com/sirupsen/logrus` &rarr; go-compose-installer

server/cmd/code-annotation/serializer/serializers.go
dpordomingo: It was quite easy to implement in our case, but I'd reconsider it for sure, and ping you when done :dagger: Thanks for being nitpicker !!!

server/cmd/code-annotation/serializer/serializers.go
dpordomingo: You may be right. I'll reconsider it and I'll ping you with the answer. Thanks for suggesting it!

server/model/models.go
dpordomingo: I just tried to follow our [specs](https://github.com/src-d/code-annotation/issues/2#issuecomment-357315069) because I was told that other log-in providers are not expected in the middle term. (We even considered using the `gh-user-name` as the `User.ID`) Anyhow, I'd consider https://github.com/src-d/code-annotation/pull/22/files#diff-c318b873cebb799af6f409abe8089672 for sure.

About storing the avatar... I'd like to confirm it with @src-d/product, because in our former web apps they requested us to do not store it, but loading it from gh (so it would be always updated) `htps://avatars2.githubusercontent.com/u/{gh-user-id}`

To do so, let us know if you can retrieve the `gh-user-id` from GH, or if we should store it in the DB.

server/cmd/code-annotation/handler/assignments.go
smacker: I would put it in handler as you posted snippet above for now. Later if we have more cases we can refactor it to service.

server/cmd/code-annotation/handler/assignments.go
smacker: could you please explain why? How are you going to use serializers beside in handlers? Because in Go usually, you export functions/types if you consider them as your API. 
And it's a good idea to keep your API as small as possible.

server/cmd/code-annotation/handler/assignments.go
smacker: I would prefer.
```go
func UpdateAssignment(model.Assignment) error
```

and in handler or where you create model:
```go
id := getId(r)
payload := decodeAssigment(r.Body)
if err := payload.Validate(); err != nil {
  // handle error
}
a, err := repository.GetAssignment(id)
if err != nil {
  // handle error
}
// could be abstracted to a function
a.Answer = payload.Answer
a.Duration = payload.Duration
// UpdateAssignment doesn't care how "a" was created
repository.UpdateAssignment(a)
```
why would I do like this? Because logic of modifying model belongs to handler. I don't think repository should know anything about payload.

```go
func UpdateAssignment(id int, answer model.Assignment) error
```
has a problem: what if `answer` has `ID` already?
The behavior will be difficult to understand.

But I'm not very strong opinionated about it. If you prefer another style I'm fine with it.

server/cmd/code-annotation/handler/assignments.go
smacker: `chi/middleware/Logger` the name of the package says you it's not supposed to be used outside of middleware. There is no problem to use middleware with logrus: https://github.com/go-chi/chi/blob/master/_examples/logging/main.go

`log15` is also fine for me, can we vote`logrus` vs`log15` then?
for go-git using standard log makes sense. it's library, not application.

server/model/models.go
dpordomingo: That values should not be exposed in the UI, never-ever.

**TL;DR**
(this answer is based in https://github.com/src-d/code-annotation/issues/2#issuecomment-357315069 but feel free to keep answering me if the conclusions below does not satisfy you)

### MTurk

In *MTurk*, the `Assignment` tracks the `Answer` of a HIT, as we do, but the `Answer` of an `Assignment` is not defined by the platform itself but by the `Requester`.

In *MTurk* the `Answer` schema is defined by the `Project` itself,
An MTurk `Answer` schema could be whatever nested thing, a quite simple could be:
`{name:string, age:int, gender:bool, desc:string, croppedImage:[]blob}`
but we don't have that concept because we were requested:
- to use a single `Project` type, whose `Answer` schema would be: `{answer:int[1-4]}`,
- to avoid unnecessary normalization;

because of that requirements, we decide by https://github.com/src-d/code-annotation/issues/2#issuecomment-357315069 :
- to have `Experiment` instead of `Project`,
- to avoid `Answer` model, and just embed it in the `Assignment`

Considering that our accepted answers are an `enum` of size 4, we just use that "strings" instead of using `int[1-4]`.
Because of that:
- the internal value of the `Assignment.Answer` should not be exposed to the UI.
- the Front should have a mapping ( DB/API &RightArrowLeftArrow; UI ) between the `enum` values used in the DB/API, and labels that the user will see.

### Supervise.ly

Supervise.ly focuses on image labeling (so they does not have `Projects` neither) and their answers are class oriented. Because of the same reason than in MTurk, the class names are always defined by the `Requester`, and they can be whatever random word or group of chars (example:  `Car`, `Dog`, `Cake`, `UFA4E4AS`...

server/cmd/code-annotation/serializer/serializers.go
smacker: I don't advocate for this particular code, just another way how to do it for you to consider: https://github.com/src-d/code-annotation/pull/22

I also find empty return approach error prone. In my previous project, I myself fixed like 5 bugs about it before we moved to non-empty return. And reviewed about 5 more fixed by other developers. I mean bugs that we found in production, not during CR.

server/model/models.go
smacker: Github return `https://avatars0.githubusercontent.com/u/406916?v=4` as avatar url. I suggest to store it.

server/cmd/code-annotation/handler/assignments.go
dpordomingo: Packages also help us to choose [good names](https://blog.golang.org/organizing-go-code#TOC_2.).

If we now care about the exposed API, it's because we're taking care of our external users, but this repo does not contain a library to be used for third parties but a repo containing a bunch of things like a web app, tools for DB handling, an API...

If we still think that a curated external API is needed here, we could [move that internal stuff to internal packages](https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU/edit) and keep the benefits of both sides.

server/cmd/code-annotation/handler/assignments.go
dpordomingo: I didn't test it yet, but [if I understood it](https://github.com/go-chi/chi/blob/master/_examples/logging/main.go#L132) the Logger can be retrieved (and used) from the inside of the handlers;
```go
type printfer interface{
  func Printf(format string, v ...interface{})
}
```
and then
```go
logger := r.Context().Value(middleware.LogEntryCtxKey).(printfer)
logger.Printf("whatever")
```

About the log library to use, I think we should consider it as a team decision, considering the company first rule; I mean, imho we should use the same logger in all our projects, and it should be the one that is "the most wanted" by engineering. What do you think about that decision @bzz ?

server/cmd/code-annotation/serializer/serializers.go
dpordomingo: I agree with the 1st paragraph.

I'm lost in "_empty return approach_". I could not find any occurrence of `empty` word in this PR. Could you give me a clue?

server/cmd/code-annotation/handler/assignments.go
bzz: Let's use Logrus for consistency.

server/model/models.go
dpordomingo: It's a business decision.

@bzz should we keep the user avatar in sync with the real one?
- **in sync**: showing whatever avatar GH shows under `htps://avatars2.githubusercontent.com/u/{gh-user-id}`
- **outdated**: showing a stored version when the user first logged in

server/cmd/code-annotation/serializer/serializers.go
smacker:
```go
func Handler(w, r) {
  if err != nil {
    write(w)
    return // empty return
  }
  write(w)
}

Comment (bzz) [reply server/cmd/code-annotation/serializer/serializers.go]: I think what Maxim means is that it's a good and less error-prone to always return something like i.e https://github.com/src-d/code-annotation/pull/22/files#diff-3e10846b2816b527e6724efedecbb964R15

File: server/model/models.go (hunk 11 above)
Comment (bzz) [reply server/model/models.go:3]: It's totally fine, either not to have avatar at all, or to show an outdated on.

On the side-note: personally, I would prefer us merging and improving already working things in subsequent small PRs rather then arguing about abstract things in single long PR.

**Comment by bzz** [comment]
@smacker could you summarize only the main points you see blocking us from merging this PR in a check-list form?

Feel free to update this one:

 - [x] change [package structure](https://github.com/src-d/code-annotation/pull/16#issuecomment-358947367) a bit
 - [x] add [Logrus](https://github.com/src-d/code-annotation/pull/16#discussion_r162639513) for logging
 - [x] reconsider how we write handlers. (kill [empty returns](https://github.com/src-d/code-annotation/pull/16#discussion_r162643097))
 - [x] better [error handing](https://github.com/src-d/code-annotation/pull/16#discussion_r162598081)

I think that would help us to decide with the order of merging things, i.e this, then #22 or other way round.

*Update*: was updated by Maxim.

@dpordomingo does it make sense to you, to fix this asap and merge then?

File: server/model/models.go (hunk 11 above)
Comment (smacker) [reply server/model/models.go:3]: 406916 is a userId, so the image itself is always fresh.

File: server/cmd/code-annotation/handler/assignments.go (hunk 6 above)
Comment (smacker) [reply server/cmd/code-annotation/handler/assignments.go]: I still don't like to have a package for only 1 particular consumer. But I can live with it.

File: server/model/models.go (hunk 12 above)
Comment (smacker) [reply server/model/models.go]: yeah. It's only internal name. But I still want to have it "correct".
Looks like `yes, no, maybe, skip` are ok if we consider it as enum 🎉 
Good to go!

File: server/cmd/code-annotation/handler/assignments.go (hunk 8 above)
Comment (dpordomingo) [reply server/cmd/code-annotation/handler/assignments.go]: That's what I was trying to find: [Choose one log library for Go and implement on all the projects](https://github.com/src-d/backlog/issues/840#issuecomment-343930191), many thanks for pointing it out !!

File: server/cmd/code-annotation/serializer/serializers.go (hunk 10 above)
Comment (dpordomingo) [reply server/cmd/code-annotation/serializer/serializers.go]: As a global assertion, I agree with it... but in the context of `chi` handlers, the expected ones must have the signature [http.HandlerFunc](https://golang.org/pkg/net/http/#HandlerFunc), that is "empty return"
In the code you linked, the empty return handler is deeper, but [it is](https://github.com/src-d/code-annotation/pull/22/files#diff-a2f70ad899ea31772f2946085a294415R39) 

If you're talking about status, errors and so on, I agree, but it's not related to the "empty return" funcs

Comment (smacker) [reply server/cmd/code-annotation/serializer/serializers.go]: it's not `chi` handlers. It's `net/http` handlers. Of course there is empty return on top level.  But I can't imagine how you can make mistake there. Everything inside won't compile without proper return.

Comment (dpordomingo) [reply server/cmd/code-annotation/serializer/serializers.go]: `chi.Router.Get(pattern string, h http.HandlerFunc)` requires a `http.HandlerFunc` (an empty return func). It does not matter what you put as `h`, that it'll be an empty return func.

**Again**: I agree on the status, errors and so on, but it's a different thing that I already confirmed it will be added to the server.

File: cli/server/server.go [hunk 13]
```diff
@@ -38,43 +38,50 @@ func main() {
 	envconfig.MustProcess("jwt", &jwtConfig)
 	jwt := service.NewJWT(jwtConfig.SigningKey)
 
-	// routing
+	// cors options
+	corsOptions := cors.Options{
+		AllowedOrigins: []string{"*"},
+		AllowedMethods: []string{"GET", "POST", "PUT", "OPTIONS"},
+		AllowedHeaders: []string{"Location", "Authorization"},
```
smacker: we also need `content-type` header

cli/server/server.go
```diff
@@ -38,43 +38,50 @@ func main() {
...
+		r.Get("/me", handler.Me(userRepo))
 	})
 
 	r.Get("/login", handler.Login(oauth))
 	r.Get("/oauth-callback", handler.OAuthCallback(oauth, jwt, userRepo, conf.UIDomain))
 
-	// protected handlers
-	r.Route("/api", func(r chi.Router) {
-		r.Use(jwt.Middleware)
+	r.Route("/experiments/{experimentId}", func(r chi.Router) {
```
smacker: should be inside api and protected

cli/server/server.go
```diff
@@ -38,43 +38,50 @@ func main() {
...
+	}
+
+	msProtected := append(ms, jwt.Middleware)
+
+	r.Route("/api", func(r chi.Router) {
+		r.Use(msProtected...)
+		r.Get("/me", handler.Me(userRepo))
 	})
 
 	r.Get("/login", handler.Login(oauth))
```
smacker: Recoverer and logger middlewares must be used for all routers

server/handler/auth.go
```diff
@@ -19,6 +18,7 @@ func Login(oAuth *service.OAuth) http.HandlerFunc {
 
 // OAuthCallback makes exchange with oauth provider, gets&creates user and redirects to index page with JWT token
 func OAuthCallback(oAuth *service.OAuth, jwt *service.JWT, userRepo *repository.Users, uiDomain string) http.HandlerFunc {
+	logger := service.NewLogger()
```
smacker: should it be a dependency? for example in tests you want to use another logger

server/middleware/appender.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/src-d/code-annotation/server/serializer"
+
+	"github.com/src-d/code-annotation/server/service"
+)
+
+func NewMiddlewaresWithAppender(ms []func(http.Handler) http.Handler) []func(http.Handler) http.Handler {
+	return append(ms, Appender)
+}
+
+func Appender(next http.Handler) http.Handler {
```
smacker: we definitely need comments here. It isn't obvious what `Appender` does

server/handler/experiments.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/go-chi/chi"
+)
+
+func GetExperimentDetails(w http.ResponseWriter, r *http.Request) {
+	//TODO: !ok
+	postponed, _ := w.(middleware.PostponedResponse)
+
+	experimentID, err := strconv.Atoi(chi.URLParam(r, "experimentId"))
+	if err != nil {
+		postponed.AddError(&serializer.Error{500, "Wrong format in experiment ID sent", err.Error()})
```
smacker: `http.StatusInternalError`

server/handler/experiments.go
```diff
@@ -0,0 +1,31 @@
...
+		return
+	}
+
+	experiment, err := repository.GetExperimentById(experimentID)
+	if err != nil {
+		postponed.AddError(&serializer.Error{404, "No experiment found", err.Error()})
```
smacker: `http.StatusNotFound`

server/handler/assignments.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/src-d/code-annotation/server/middleware"
+	"github.com/src-d/code-annotation/server/repository"
+	"github.com/src-d/code-annotation/server/serializer"
+
+	"github.com/go-chi/chi"
+)
+
+func GetAssignmentsForUserExperiment(w http.ResponseWriter, r *http.Request) {
+	//TODO: !ok
+	postponed, _ := w.(middleware.PostponedResponse)
```
smacker: what is the point to "postpone" response? I checked handlers and I don't see where you might need it. Every time you `return` you can create normal response from the line just above that return.

server/middleware/appender.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+		defer p.Close()
+
+		next.ServeHTTP(p, r)
+
+		p.flush()
+	})
+}
+
+type postponed struct {
```
smacker: wrapping `http.ResponseWriter` isn't such an easy task. You can read about it more [here](https://github.com/felixge/httpsnoop#why-this-package-exists) for example.

server/handler/auth.go
dpordomingo: I started making the logger injectable/configurable, so you could define which one to use in different scenarios.
But I rollbacked to remove initial complexity.
Do we want that for this first iteration?

smacker: where is a complexity to change 
```go
func OAuthCallback() {
  logger := service.NewLogger()
}
```
to
```go
func OAuthCallback(logger *logrus.FieldLogger) {
}
```
?

server/middleware/appender.go
dpordomingo: Yes, you're right.
And the name is not clear at all.
Options:
- now,
- merge and document.

Anyhow, I planned to add it to the login/callback so all errors, logging... will be homogeneous

server/handler/experiments.go
dpordomingo: Yes, I planned to review all errors codes and texts to use constants and make them more clear.
Should I do it before merging? Just say so :dancer:

server/handler/experiments.go
dpordomingo: _ditto_

server/handler/assignments.go
dpordomingo: the postponed Responses are the way to avoid writing things randomly; that way the handlers push the results to the postponed Response, and at the end of the flow it is processed and properly serialized, the errors are checked...

server/middleware/appender.go
smacker: I think we need to document it now. Because I don't understand it completely even for review.
Like what would happen if I use `w` as a normal writer. Looks like it will just fail even if you implement `http.ResponseWriter` methods instead of current panic.
Is it normal behavior? Is it a bug?
Without documentation I don't know.

server/handler/experiments.go
smacker: it's a very simple change. Why postpone?

server/handler/assignments.go
smacker: you are talking about transformation. Not postponing.

Postponing is something like
```go
func handler(w) {
  writeSomethingTo(w)
  // do long computation maybe in goroutine
  writeMoreTo(w)
}
```

server/middleware/appender.go
dpordomingo: I tried to emulate the same flow that [chi/middleware/compress](https://github.com/go-chi/chi/blob/master/middleware/compress.go) and [chi/middleware/logger](https://github.com/go-chi/chi/blob/master/middleware/logger.go) follows. Both wrap it to ensure the response is created in a controlled and homogeneous way.

Thanks for the resource, I'll read a bit more to consider unhandled circumstances.

server/handler/auth.go
dpordomingo: It is now:
```go
func OAuthCallback(oAuth *service.OAuth, jwt *service.JWT, userRepo *repository.Users, uiDomain string) http.HandlerFunc {
```
I didn't want to add more params to some handlers, but I can do so for sure.
I saw that chi/logger uses the context to store the logger, what could be another approach.
Which one would be your option?
- a: one parammeter more,
- b: using the context as chi does

server/handler/experiments.go
dpordomingo: I didn't want 500 for everything because there could be any other with more sense. But yes, that enhanced can be postponed too.
Adding `http` constants now

server/handler/assignments.go
dpordomingo: The Response is written after the handler is run

server/handler/auth.go
smacker: the easiest way to just add one more param. The better way is to convert it to structure.
for logger it's okay to store it in context (but you will need a helper, which returns "default" logger if there is no logger in context), especially if you want to have request scoped logger (and I want).

But remember `Context.Value should inform, not control`. Here is a good article about it: https://medium.com/@cep21/how-to-correctly-use-context-context-in-go-1-7-8f2c0fafdf39

server/middleware/appender.go
dpordomingo: If you're using that middleware, the w you receive won't accept writes.
I'm writing a couple of lines with doc; I'll ping you in a few mins when ready :D

server/handler/auth.go
dpordomingo: so `a`, and lets proceed with `b` in the next iteration

server/handler/assignments.go
smacker: I get it. I'm saying the name is confusing. If you want to keep using this approach (I don't) the name like `Responser` is much better. Because it's interface and consumer doesn't need to know how it works internally (does it postpone anything or not)

server/middleware/appender.go
smacker: then it's completely incorrect code.
handler accepts `http.ResponseWriter`. But it isn't `http.ResponseWriter` anymore.

cli/server/server.go
dpordomingo: addressed by 28c54c4

cli/server/server.go
dpordomingo: addressed by 28c54c4

cli/server/server.go
dpordomingo: addressed by 28c54c4

server/handler/auth.go
dpordomingo: addressed by 28c54c4

server/middleware/appender.go
dpordomingo: functionality dropped

server/handler/assignments.go
dpordomingo: functionality dropped

server/middleware/appender.go
dpordomingo: functionality dropped

dpordomingo:
Done, here is the summary:
- change package structure a bit as agreed https://github.com/src-d/code-annotation/pull/16#issuecomment-358947367
- add Logrus for logging https://github.com/src-d/code-annotation/pull/16/commits/2400963bd9ae789a8829aa3956f6852eb776a38f
- reconsider how we write handlers and error handing 28c54c4

I'll review the constants for status because there could be some missing

dpordomingo:
**PS:**
Many thanks to all of you for being nitpicking :dagger:

cli/server/server.go
```diff
@@ -0,0 +1,97 @@
...
+	// loger
+	logger := service.NewLogger()
+
+	r := chi.NewRouter()
+
+	ms := []func(http.Handler) http.Handler{
+		middleware.Recoverer,
+		cors.New(corsOptions).Handler,
+		middleware.RequestLogger(&middleware.DefaultLogFormatter{Logger: logger}),
+	}
```
smacker: better write as in documentation:
```go
r := chi.NewRouter()
r.Use(middleware.Recoverer)
r.Use(cors.New(corsOptions).Handler)
r.Use(middleware.RequestLogger(&middleware.DefaultLogFormatter{Logger: logger}))

r.Get("/login", GetHandler(handler.Login(oauth)))
r.Get("/oauth-callback", GetHandler(handler.OAuthCallback(oauth, jwt, userRepo, conf.UIDomain, logger)))

r.Route("/api", func(r chi.Router) {
  r.Use(jwt.Middleware)
  ...
})
```

not because current code is wrong, but because it isn't that easy to understand.
Like I need to scan all code below this line for `ms`.

If you want to exclude static (though I don't see why, I think it should be logged also and `FrontendStatics` may panic also). You can wrap it in another `r.Group`. But `r.Use`, `r.Use` is much more readable than array.

server/handler/assignments.go
```diff
@@ -0,0 +1,71 @@
...
+	"strconv"
+
+	"github.com/src-d/code-annotation/server/repository"
+	"github.com/src-d/code-annotation/server/serializer"
+
+	"github.com/go-chi/chi"
+)
+
+func GetAssignmentsForUserExperiment() RequestProcessFunc {
+	return func(r *http.Request) (resp *serializer.Response, errs []*serializer.Error) {
```
smacker, line 16: https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

also you lose all the profit of go type checking.

consider code:
```go
func(r *http.Request) (resp *serializer.Response, errs []*serializer.Error) {
  ...a lot of code...
  if somecheck {
    resp = postResponse()
  }
  ...a lot of code...
  resp = getResponse()
  return
}
```
code above compiles and produces wrong response.

this signature and always return new value avoids the problem:
```go
func(r *http.Request) (*serializer.Response, []*serializer.Error) {
```

server/service/jwt.go
```diff
@@ -8,7 +8,7 @@ import (
 
 	"github.com/src-d/code-annotation/server/model"
 
-	jwt "github.com/dgrijalva/jwt-go"
+	"github.com/dgrijalva/jwt-go"
```
smacker, line 11: I'm sure my editor will fix it back as soon as I open this file.
I'm too lazy to look for a link right now, but it's good practice to use named import if name of package != import path.
With jwt-go it's not that bad. But sometimes names are much different and it's just difficult to understand where this package came from.

server/handler/auth.go
```diff
@@ -6,48 +6,55 @@ import (
...
 )
 
 // Login handler redirects user to oauth provider
-func Login(oAuth *service.OAuth) http.HandlerFunc {
-	return func(w http.ResponseWriter, r *http.Request) {
+func Login(oAuth *service.OAuth) RequestProcessFunc {
+	return func(r *http.Request) (resp *serializer.Response, errs []*serializer.Error) {
 		url := oAuth.MakeAuthURL()
-		http.Redirect(w, r, url, http.StatusTemporaryRedirect)
+		resp = serializer.NewTemporalRedirect(url)
```
smacker, line 17: what is the point of using multiple layers of logic for it if you actually just do `http.Redirect` anyway?

server/handler/renderer.go
```diff
@@ -0,0 +1,44 @@
...
+	//TODO: Responses and Redirects should implement Serializable{Serialize(w)}
+	if response.RedirectURL != "" {
+		http.Redirect(w, r, response.RedirectURL, response.Status)
+		return
+	}
+
+	status := http.StatusOK
+
+	if len(errs) > 0 {
+		status = http.StatusInternalServerError
```
smacker, line 29: something is wrong here.
if handler returns error `not found` the status should be `not found` not `internal server error`

server/handler/renderer.go
```diff
@@ -0,0 +1,44 @@
...
+		response = &serializer.Response{Status: status, Errors: errs}
+		for _, e := range errs {
+			//TODO: Not all errors should be logged as .Error()
```
smacker, line 32: 4xx statuses are client errors, you don't need to log them
5xx statuses are server errors, we need to log them

server/handler/static_files.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"net/http"
+	"os"
+)
+
+func FrontendStatics(w http.ResponseWriter, r *http.Request) {
+	filepath := "build" + r.URL.Path
+	if _, err := os.Stat(filepath); err == nil {
+		http.ServeFile(w, r, filepath)
+	}
```
smacker: and here we go. empty return = mistake.

it was easy to spot in a handler with 5 lines of code.
But in with more code it's' more difficult.

That's why I'm so aggressively against of empty return when we don't have to use them.

server/handler/assignments.go
```diff
@@ -0,0 +1,71 @@
...
+		experimentID, err := strconv.Atoi(chi.URLParam(r, "experimentId"))
+		if err != nil {
+			errs = append(errs, &serializer.Error{500, "Wrong format in experiment ID sent", err.Error()})
```
smacker, line 19: sorry I don't remember in which document it's described. But `go vet` won't allow this code with warning like `composite literal uses unkeyed fields` https://golang.org/cmd/vet/#hdr-Unkeyed_composite_literals

And it makes a lot of sense!
If you change the order of fields (last 2) in your struct - code still compiles but works incorrectly.
Also it will be very painful to add a new optional field.

server/repository/repositories.go
```diff
@@ -0,0 +1,68 @@
...
+		&model.Assignment{ID: 5, UserID: userID, PairID: 1, ExperimentID: experimentID},
+		&model.Assignment{ID: 6, UserID: userID, PairID: 2, ExperimentID: experimentID},
+		&model.Assignment{ID: 7, UserID: userID, PairID: 1, ExperimentID: experimentID},
+		&model.Assignment{ID: 8, UserID: userID, PairID: 2, ExperimentID: experimentID},
+	}
+
+	return assignments, nil
+}
+
+func UpdateAssignment(assignmentId int, assignment *model.Assignment) error {
```
smacker: looks like you missed my comment before.
What behavior should I expect from this method if `assignmentId` != `assignment.ID`?
I don't have a correct answer.
That's why I suggested to change func signature.

server/serializer/serializers.go
```diff
@@ -0,0 +1,105 @@
...
+	return &Response{Status: http.StatusTemporaryRedirect, RedirectURL: URL}
+}
+
+type ExperimentResponse struct {
+	ID          int    `json:"id"`
+	Name        string `json:"name"`
+	Description string `json:"description"`
+}
+
+func NewExperimentResponse(e *model.Experiment) *Response {
```
smacker, line 47: you just broke name convention.
If a package has public `ExperimentResponse` structure, `NewExperimentResponse` should return it. Not `Response`.
Returning`Response` will make sense if `ExperimentResponse` is private.

server/repository/repositories.go
dpordomingo: From the specs https://github.com/src-d/code-annotation/issues/3#issuecomment-357760945, the assignment request comes with no ID, so [it is not considered](https://github.com/src-d/code-annotation/pull/16/files/28c54c4e696585b8342c42af4e446eb5e7b8959a#diff-1731f4c3708fc9f3243464af602a7b16R58)
Same happens with `answer.PairID`: it is ignored.
It will be ensured in the tests too.

Anyhow, I'll rollback to 
```
func UpdateAssignment(assignmentId int, answer string, duration int) error
```
That is more clear and guarantee our requirements by design

smacker: we need to add `id` to an assignment. Because FE needs to update it. RESTful way to do it is to call `PUT .../assignments/{id}`

server/handler/static_files.go
dpordomingo: The problem was caused because the `return` from the original sourced dissapeared (:unamused:) when the original code was moved to the handler package.

I don't see how could it be avoided by a qualified return; example, given the following code:
```go
func () string {
    if (cond) {
        doSomething()
        return "true"
    }
    return "false"
}
```
the one above will compile too, and has the same problem.

```go
func () string {
    if (cond) {
        doSomething()
    }
    return "false"
}
```

smacker:
```go
func FrontendStatics(w http.ResponseWriter, r *http.Request) {
    http.ServeFile(w, r, func() string {
        filepath := "build" + r.URL.Path
        if _, err := os.Stat(filepath); err == nil {
            // you can't write filepath without return, it won't compile
            return filepath
        }
        return "build/index.html"
    }())
}
```

but I don't suggest to change this handler. It's small, so no need to do it more complex. But for other handlers it makes sense.

P.S. Of course you can always make a mistake, just the chance is lower with non-empty returns.

dpordomingo:
superseded by https://github.com/src-d/code-annotation/pull/37

PR #23, opened 2018-01-19
Tool to convert file pairs in json to an SQLite DB

carlosms:
This is an internal tool for the moment. Do not expect production-ready documentation, maintenance, or tests.


bzz (approved):
LGMT

smacker (approved):
as long as it does the job it's good for me!

PR #28, opened 2018-01-23
Add option to use postgres backend in import and output

carlosms:
This commit adds the functionality for the import tool (#4).
A pending commit will add the same options to the export tool (#5)

The basic change is that arguments can now be:
sqlite:///path/to/db.db
postgresql://user:password@host:port


server/dbutil/dbutil.go
```diff
@@ -49,14 +79,54 @@ const insertFilePairs = `INSERT INTO file_pairs (
 		score, diff, experiment_id ) VALUES
 		($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15)`
 
+// Open returns a DB from the connection string.
+// With checkExisting it will fail if the DB does not exist
+func Open(connection string, checkExisting bool) (DB, error) {
+	sqliteReg := regexp.MustCompile(`^sqlite://(.+)$`)
+	psReg := regexp.MustCompile(`^postgres(ql)?:.+$`)
```
smacker: you are checking only for prefix, right?
Would it be easier to use `string.HasPrefix`?
and for `sqlite` to remove prefix you can use `string.TrimPrefix`

just a suggestion, up to you.

smacker (approved):
I don't like much being too smart and trying to invent abstraction on top of different databases when people solved this problem multiple times already and better than we.

But the rest looks ok. Approve.

carlosms:
> I don't like much being too smart and trying to invent abstraction on top of different databases when people solved this problem multiple times already and better than we.

I have to agree with you, this may not be most elegant approach. The tool started simple enough that I just went with pure sql. Then postgres had to be added, and the code was already there for sqlite so...

server/dbutil/dbutil.go
carlosms: I like to use regexp if there is any excuse :D

cli/export/export.go
```diff
@@ -1,27 +1,40 @@
 /*
 Tool to export annotation results from the internal DB to an output DB.
+It does a simple copy&paste of the internal DB. The annotation results are
+stored in the assignments table.
 
-Usage: export <path-to-origin.db> <path-to-destination.db>
+Usage: export DSN DSN
```
bzz: Is DSN better then DB_URL ?

Also related - the requirement on the UI level for export is: 
 - input: url to point to inernal DB
 - output: path in the filesystem to sqlight DB

And viasurvesa for import.

I believe it has advantage of simplicity for this app the user, at the same time agility for its developers (choose any DB)

Could you please update it?

smacker: imo DSN is better. It's standard name for it. If I read DB_URL, I'm not sure what format should be used.
For other points 👍

bzz: 👍  a good argument for using DSN for DB part also might be posting something like https://en.wikipedia.org/wiki/Data_source_name :)

smacker: there even a wiki page for it! I just saw it across documentation of different drivers & tools.

carlosms: @bzz:

> input: url to point to inernal DB
> output: path in the filesystem to sqlight DB

Please confirm: you want the output argument to accept only sqlite:// strings, or do you want it to ask for a path? i.e.

1. `export postgresql://user:password@host:port sqlite:///path/to/db.db`
2. ` export postgresql://user:password@host:port /path/to/db.db`

bzz: 2 in my opinion follows the right contract.
Import would be `import /path/to/db.db postgresql://user:password@host:port`

I guess this PR could also take care of doc update, after #29 is merged.

bzz (approved):
LGTM

server/dbutil/dbutil.go
```diff
@@ -6,23 +6,47 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"regexp"
+	"strings"
 
+	_ "github.com/lib/pq"
```
bzz, line 12: It looks like this PR adds a new dependency library.

As https://github.com/src-d/code-annotation/pull/33 is already merged, this means import/export tools should also use `dep` and thus adding a dependency would include modification of the .toml file, right?

I guess that might also imply updates for https://github.com/src-d/code-annotation/pull/29 on how import/export tools are built.

bzz (changes requested):

server/dbutil/dbutil.go
carlosms: makes sense. I'll rebase from master and update dep.

bzz (approved):

Issue #13, opened 2018-01-18
Backend: introduce dependency management


bzz:
`go dep` this is a standard way srcd is adopting for golang project for avoiding problems with dependency management in Go.

smacker:
blocked: we need backend first

PR #29, opened 2018-01-24
Documentation improvements

smacker:
Fixes: https://github.com/src-d/code-annotation/issues/17

PR #33, opened 2018-01-25
use go dep for dependency management

smacker:
Fixes: https://github.com/src-d/code-annotation/issues/13


carlosms (approved):
LGTM

bzz (approved):
LGTM.

One request, if possible, is there any chance we could merge #29 first and then, in here, also update CONTRIBUTING.md, pointing out that `dep` is used on a backend for dependency management, with a link to something like https://golang.github.io/dep/docs/daily-dep.html ?

Asking as this would help to assimilate the knowledge, necessary to adopt `dep` for other projects.

mcuadros (changes requested):
vendor folder should be upload, otherwise is anti-pattern in go

smacker:
@mcuadros it's not true if we use dep: https://golang.github.io/dep/docs/FAQ.html#best-practices

smacker:
I don't think it's very right decision for this project because go here isn't the main part and in my opinion, it's better to follow the same patterns for all parts of the project. But vendor dir is added.

smacker:
vendor directory is added as Maximo requested. I need this PR to merge another one on top of is so, I'm merging (there are 2 approvals already includes maintainer)

mcuadros:
Why was this PR merged? With a pending request?

smacker:
the request is done.
we need to move faster for code-annotation with merging because we have hard deadlines.

smacker:
oh. and if you are talking about @dpordomingo review, we have an agreement to add everyone of the team in CR. But merge when there is approve of maintainer + 1 more.
If there will be comments later they will go to separate PR.

mcuadros:
Who should confirm if the changes were addressed is the author of the request, since now that the vendor is there the rule `godep`  is not required anymore.

smacker:
godep is still needed for development. Please feel free to add any comments. They will be resolved in additional PRs.

mcuadros:
Also will be nice upload only the package that are you using running first `dep prune` and not all the files

smacker:
Thanks for the suggestion. I created an issue: https://github.com/src-d/code-annotation/issues/39

PR #35, opened 2018-01-26
UI update

### LINKED ISSUE #19: Design polishing
The part about annotation is done by product team. We need to:

1. Align frontend with latest wireframes
2. Refactor FE to use CSS Grid layout
3. Deploy branch on staging for @ricardobaeta 
4. Collect feedback from @ricardobaeta (as css/less)
5. Apply feedback

smacker:
Partly fixes https://github.com/src-d/code-annotation/issues/19

Grid thing will be done in a separate PR.

### DIFF
--- a/src/pages/Experiment.js
+++ b/src/pages/Experiment.js
@@ -65,7 +65,7 @@ class Experiment extends Component {
   }
 
   renderMain() {
-    const { error, loading, percent } = this.props;
+    const { error, loading, name, description, percent } = this.props;
 
     if (error) {
       return (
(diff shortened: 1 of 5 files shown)


bzz:
👍 In case of UI changes, in order to speed up the review it might help to add screenshot as well.

smacker:
there are not only visible parts has been changed. But for example confirmation popup was removed. We should use staging for it. But we don't have it right now.

carlosms (approved):
Works in my machine

bzz (approved):
LGTM.

Thank you for double-checking, @carlosms !

PR #37, opened 2018-01-26
Prepare Router skeleton

dpordomingo:
supersedes https://github.com/src-d/code-annotation/pull/16
depends on https://github.com/src-d/code-annotation/pull/43
partially solves #3

This PR prepares the skeleton of the backend api serving mocks, using the requirements defined by https://github.com/src-d/code-annotation/issues/3, and following the agreements on the implementations as we talked irl

If approved, the things that will be done in the next PRs will be:
- using DB stuff, 
- tests,
- using user from Auth


server/handler/render.go
```diff
@@ -2,49 +2,62 @@ package handler
...
 }
 
-func respInternalErr() response {
-	return respErr(http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError))
-}
-
-type errObj struct {
-	Title string `json:"title"`
-}
+// TODO: it should be private
```
smacker: imo, don't need to be private.

carlosms (approved):

server/serializer/serializers.go
```diff
@@ -0,0 +1,124 @@
...
+// StatusCode returns the Status of the httpError
+func (e httpError) StatusCode() int {
+	return e.Status
+}
+
+// Error returns the string content of the httpError
+func (e httpError) Error() string {
+	if msg := e.Title; msg != "" {
+		return msg
+	} else if msg = http.StatusText(e.Status); msg != "" {
```
smacker: for the future: you don't need to use `else` if you `return` before

smacker (approved):
I saw few issues (for example about logger), but we can fix it later. Good to go!!!!!

server/handler/render.go
dpordomingo: I made it private because nobody from the outside of the package should write in the response; should it?

server/serializer/serializers.go
dpordomingo: It was done to reuse the `msg` implicit definition made by the first `if`.
Anyhow, there is no necessity of reusing it, so I'll separate it into two different `if`, and redefine the `msg`

server/handler/render.go
smacker: I can easily imagine when we would use it.
For example, we would create an endpoint to download diff or file.
```go
func GetFile(r, w) {
  name = getFileName(r)
  f, err := getFile(name)
  if err != nil {
    Write(w, Error{"file with name %s doesn't exist", name})
  }
  http.ServeContent(w, f)
}
```

server/serializer/serializers.go
dpordomingo: done by https://github.com/src-d/code-annotation/pull/37/commits/39208c342200a46bde038d97edcc4d8b68f05975

server/handler/render.go
dpordomingo: but... that endpoint should be a handler, right?
If so, it can use `handler.write` too, as done by `handler.OauthCallback`
https://github.com/src-d/code-annotation/pull/37/commits/76ed47205d7755e30d31fbbdbc26a7381e4de352#diff-3ad2ebeb7ccf861129aa87b26fa1f1d8R31

smacker: ahhhh. you moved it from serializer to handler. then all good!

dpordomingo:
Sorry @smacker and @carlosms for making you review this again, but I had to address the comments:
- Convert some internals to private things:
  - 821153d and : `handler.assignmentRequest`, 
  - 76ed472 `handler.write` and `serializer.countResponse`
- Cosmetics:
  - 39208c3 Separate if/return/else as suggested by @smacker
  - c97bdb7 Cosmetics: Run go-vet. Rename id>ID and unkeyed structs

Once you approve that last changes, I'll squash and merge. Thanks

bzz (approved):
LGTM.

Awesome job @dpordomingo!

Shall we rebase on latest master and merge?

carlosms:
LGTM

dpordomingo:
I rebased 9983b3f but I couldn't merge because this PR depends on https://github.com/src-d/code-annotation/pull/43

I also added 4f445cd with `github.com/rs/coors` dependency as required by [`./server/router.go`](https://github.com/src-d/code-annotation/pull/37/files#diff-e9d5659a23b8025687444882c47e2188R12)

PR #42, opened 2018-01-29
[WIP] Bootstrap grid

smacker:
For now this PR exists only to ask for ideas


src/components/Experiment/Footer.js
```diff
@@ -24,26 +23,30 @@ function Footer({
...
-          <Button onClick={finish}>Finish</Button>
-          <Button bsStyle="info" onClick={skip}>
-            Skip
-          </Button>
-        </div>
-      </div>
-    </div>
+        </Col>
+        <Col xs={6} className="text-center">
+          <ButtonToolbar style={{ display: 'inline-block' }}>
```
dpordomingo, line 28: Why is this inline style really needed?

dpordomingo: I mean... why not doing it in the css side?

smacker: the easiest way to center ButtonToolbar I found. Parent with `text-center` + `inline-block`.
any ideas?

src/pages/Experiment.less
```diff
@@ -38,20 +27,27 @@
...
+        height: 100%;
+
+        // must be sum of everything on top + footer
+        // page-header: 50px
+        // experiment-header: 23px
+        // margins: 20px * 2 + 10px
+        // footer: 46+10px
+        // borders: 2px
+        // 50 + 23 + 40 + 56 + 10 + 2 = 181px
+        padding-bottom: 181px;
```
dpordomingo, line 44: What if all that things are variables, so nothing would be broken if some height changes?

dpordomingo (commented):
I see it kind of ok.

As I understood, you removed the custom positioning definitions and just used bootstrap-react stuff, and I think it's ok.

If the PR is approved, (a) is it being merged as it is? or (b) do you have any other pending changes?
If (a), I'd try to confirm (before merging) there are no other hardcoded positions in our components.

smacker:
@dpordomingo it's WIP. Only one page was changed. As I mentioned in the description this PR exists only for ideas purposes. I'll create a new one later.

src/pages/Experiment.less
smacker: there is no fixed height for blocks around.
For example height of footer is actually is the height of buttons. Even if you define the height for the footer as soon as you make buttons bigger (or add margin to them or whatever) it will get broken.

src/components/Experiment/Footer.js
smacker: will move to css, no problem here

PR #43, opened 2018-01-29
Add gorilla dependency

dpordomingo:
blocks https://github.com/src-d/code-annotation/pull/37

ptal @src-d/applications 

Add gorilla dependency as already required by [`./server/service/oauth.go`](https://github.com/src-d/code-annotation/blob/master/server/service/oauth.go#L11)


bzz (approved):
Not sure why it is not part of #37 but LGMT.

Let's merge both asap.

dpordomingo:
@bzz it was not added in that PR because gorilla dependency was not introduced by that one

Issue #41, opened 2018-01-29
Backend: DB code


carlosms:
This issue will add code to write data to the DB, since the current code uses mock data.

carlosms:
PR merged. The only point left to discuss is explained in a new issue, https://github.com/src-d/code-annotation/issues/66

PR #52, opened 2018-01-31
Use DB in the backend

carlosms:
Implements #41.

Please take a look @dpordomingo, @smacker.

TODOs:

- [X] Update README with new env variable (done in e23160c)
- [x] Check if DB strings need to be escaped


cli/server/server.go
```diff
@@ -34,12 +42,9 @@ func main() {
 	envconfig.MustProcess("jwt", &jwtConfig)
 	jwt := service.NewJWT(jwtConfig.SigningKey)
 
-	// loger
-	logger := service.NewLogger()
-
 	// start the router
-	router := server.Router(logger, jwt, oauth, conf.UIDomain, userRepo, "build")
+	router := server.Router(logger, jwt, oauth, conf.UIDomain, db.SQLDB, "build")
```
smacker: does it look weird to pass DB connection to a router only for me?
(no changes required, just asking)

server/dbutil/dbutil.go
```diff
@@ -26,13 +26,13 @@ const (
 
 // DB groups a sql.DB and the driver used to initialize it
 type DB struct {
-	sqlDB  *sql.DB
+	SQLDB  *sql.DB
```
smacker: what do you think about composition here?
It would allow not write `db.SQLDB.method` everywhere, just `db.method`

cli/server/server.go
carlosms: I'm ok with moving the repositories initialization back to the server.go file. I just did that to avoid having lots of arguments here. But of course we could always store all the repos in a single struct...

Let me know if I should change this.

server/handler/assignments.go
```diff
@@ -45,16 +49,22 @@ type assignmentRequest struct {
 }
 
 // SaveAssignment returns a function that saves the user answers as passed in the body request
-func SaveAssignment() RequestProcessFunc {
+func SaveAssignment(repo *repository.Assignments) RequestProcessFunc {
 	return func(r *http.Request) (*serializer.Response, error) {
-		requestedAssignmentID := chi.URLParam(r, "assignmentId")
-		assignmentID, err := strconv.Atoi(requestedAssignmentID)
+		requestedPairID := chi.URLParam(r, "pairId")
+		pairID, err := strconv.Atoi(requestedPairID)
```
smacker: I see we do it often. Just an idea for the future:

```go
func URLParamInt(r, key) (int, error) {
  str := chi.URLParam(r, "pairId")
  return strconv.Atoi(str)
}
```

server/handler/auth.go
```diff
@@ -34,21 +35,30 @@ func OAuthCallback(
...
-			logger.Errorf("can't create user: %s", err)
+		user, err := userRepo.Get(ghUser.Login)
+		if err != nil {
+			logger.Error(err)
 			write(w, r, serializer.NewEmptyResponse(), err)
 			return
 		}
 
+		if user == nil {
+			user, err = userRepo.Create(ghUser.Login, ghUser.Username, ghUser.AvatarURL, model.Requester)
```
smacker: I noticed this pattern with passing fields of structs a few times already.
No changes are required now, but I prefer to pass a model to a repository.
It makes life easier when the model starts growing.

server/repository/assignments.go
```diff
@@ -0,0 +1,145 @@
...
+		return nil, fmt.Errorf("DB error: %v", err)
+	}
+
+	rows, err := repo.DB.Query(
+		"SELECT id FROM file_pairs WHERE experiment_id=$1", experimentID)
+	if err != nil {
+		return nil, fmt.Errorf("Error getting file_pairs from the DB: %v", err)
+	}
+
+	answer := ""
```
smacker: according to https://github.com/src-d/code-annotation/issues/2 it should be `nil`. Frontend also expects `nil` there.

server/handler/auth.go
carlosms: Makes sense. But the problem, in this specific case, is that there is a field of the User model that is not up to you, the ID will be chosen by the DB.

You could create a User model and leave the ID as 0, but that can lead to bugs in my opinion... you may be tempted to continue using the `User` object you created. In the current code it is clear for the caller that the complete `User` object will be provided by `Create`.

Any other opinions?

server/repository/file_pairs.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"database/sql"
+	"fmt"
+
+	"github.com/src-d/code-annotation/server/model"
+)
+
+// FilePairs repository
+type FilePairs struct {
+	DB *sql.DB
```
smacker: maybe it should be private?

cli/server/server.go
smacker: no, please don't. It simplifies the code.

server/handler/assignments.go
```diff
@@ -28,9 +28,13 @@ func GetAssignmentsForUserExperiment() RequestProcessFunc {
 		}
 
 		userID := service.GetUserID(r.Context())
-		assignments, err := repository.GetAssignmentsFor(userID, experimentID)
+		if userID == 0 {
```
smacker: we do it often. How about moving it to a helper?
```
func checkUserInCtx(ctx) (id, error)
```

server/repository/assignments.go
carlosms: We could use a null value in the DB, but isn't it going to end up as an empty string when we read it back?
https://github.com/src-d/code-annotation/pull/52/files#diff-b01fac8859e3c130c7e1a433072473edR24

smacker: no if you use [sql.NullString](https://golang.org/pkg/database/sql/#NullString)

carlosms: Ok... what about in serializers.go, how does the final json field look like?

smacker: it should be `*string`

server/handler/auth.go
smacker: Sorry I don't get the problem with ID.

```
u := User{login,name,...}
err := repo.save(&u)
fmt.Println(u.ID) // prints user id
```

keep id as zero value for objects that aren't saved yet is totally normal.

You can see in for example in [most popular ORM for go](https://github.com/jinzhu/gorm)

dpordomingo:
@carlosms please, ping me once you address the Maxim feedback to perform a second review :D

dpordomingo:
@carlosms your work looks great,
please, ping me once you address the Maxim feedback to perform a second review :D

server/repository/file_pairs.go
carlosms: DB is initialized in router.go. Or should it be a private field with a setter method?

server/dbutil/dbutil.go
carlosms: Done in 77f23a1

server/handler/assignments.go
carlosms: done in 01f1cbc

server/handler/auth.go
carlosms: done in b09114c

server/handler/assignments.go
carlosms: done in c9d26ae

server/repository/assignments.go
carlosms: done in 07e2f9c

server/repository/assignments.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	insert, err := tx.Prepare("INSERT INTO assignments VALUES ($1, $2, $3, $4, $5)")
+	if err != nil {
+		return nil, fmt.Errorf("DB error: %v", err)
+	}
+
+	rows, err := repo.DB.Query(
```
smacker: defer rows.Close()

server/repository/assignments.go
```diff
@@ -0,0 +1,144 @@
...
+func (repo *Assignments) Get(userID, pairID int) (*model.Assignment, error) {
+	return repo.getWithQuery(repo.DB.QueryRow(
+		"SELECT * FROM assignments WHERE user_id=$1 AND pair_id=$2",
+		userID, pairID))
+}
+
+// GetAll returns all the Assignments for the given user and experiment IDs.
+// Returns an ErrNoAssignmentsInitialized if they do not exist yet
+func (repo *Assignments) GetAll(userID, experimentID int) ([]*model.Assignment, error) {
+	rows, err := repo.DB.Query(
```
smacker: defer rows.Close()

server/repository/file_pairs.go
smacker: usually, people do it with a constructor

smacker (changes requested):

server/repository/assignments.go
carlosms: done in e3886d2

server/handler/render.go
```diff
@@ -64,3 +66,19 @@ func write(w http.ResponseWriter, r *http.Request, response *serializer.Response
 	w.WriteHeader(statusCode)
 	w.Write(content)
 }
+
+// URLParamInt returns the url parameter from a http.Request object. If the
```
dpordomingo: s/URLParamInt/urlParamInt
Anyhow, private functions use to be undocummented.

server/handler/render.go
```diff
@@ -64,3 +66,19 @@ func write(w http.ResponseWriter, r *http.Request, response *serializer.Response
...
+// param cannot be converted to int, it returns a serializer.NewHTTPError
+func urlParamInt(r *http.Request, key string) (val int, err error) {
+	str := chi.URLParam(r, key)
+	val, err = strconv.Atoi(str)
+
```
dpordomingo: We're trying to avoid named returns.
The code below this comment can be replaced by:
```go
if err != nil {
    err = serializer.NewHTTPError(
        http.StatusBadRequest,
        fmt.Sprintf("Wrong format for URL parameter %q; received %q", key, str)
    )
}

return val, err;
```

It is only needed to change the returned because `val` is already `0` when an error occurs.

server/router.go
```diff
@@ -46,15 +53,15 @@ func Router(
...
 
-			r.Get("/", handler.Get(handler.GetExperimentDetails()))
+			r.Get("/", handler.Get(handler.GetExperimentDetails(experimentRepo)))
 
 			r.Route("/assignments", func(r chi.Router) {
 
-				r.Get("/", handler.Get(handler.GetAssignmentsForUserExperiment()))
-				r.Put("/{assignmentId}", handler.Get(handler.SaveAssignment()))
+				r.Get("/", handler.Get(handler.GetAssignmentsForUserExperiment(assignmentRepo)))
+				r.Put("/{pairId}", handler.Get(handler.SaveAssignment(assignmentRepo)))
```
dpordomingo: It is not saved a `FilePair` but an `Assignment`, that's because it should be:
`s/pairId/assignmentId`

server/handler/assignments.go
```diff
@@ -45,14 +43,16 @@ type assignmentRequest struct {
 }
 
 // SaveAssignment returns a function that saves the user answers as passed in the body request
-func SaveAssignment() RequestProcessFunc {
+func SaveAssignment(repo *repository.Assignments) RequestProcessFunc {
 	return func(r *http.Request) (*serializer.Response, error) {
-		requestedAssignmentID := chi.URLParam(r, "assignmentId")
-		assignmentID, err := strconv.Atoi(requestedAssignmentID)
+		pairID, err := urlParamInt(r, "pairId")
```
dpordomingo: It is not saved a `FilePair` but an `Assignment`, that's because it should be:
`s/pairId/assignmentId`

server/handler/assignments.go
```diff
@@ -66,7 +66,8 @@ func SaveAssignment() RequestProcessFunc {
 			return nil, fmt.Errorf("payload could not be read")
 		}
 
-		if err := repository.UpdateAssignment(assignmentID, assignmentRequest.Answer, assignmentRequest.Duration); err != nil {
+		err = repo.Update(userID, pairID, assignmentRequest.Answer, assignmentRequest.Duration)
```
dpordomingo: It is not saved a `FilePair` but an `Assignment`, that's because it should be:
`s/pairId/assignmentId`
>_We considered that the same `FilePair` can be assigned to the same `User` many times in different experiments, so it can not be `PRIMARY KEY`_

server/model/models.go
```diff
@@ -16,30 +18,33 @@ type Experiment struct {
...
 	Duration     int
 }
 
-// FilePairs are the answers that needs to be responsed
-type FilePairs struct {
+// FilePair represents the pairs of files to annotate
+type FilePair struct {
 	ID           int
-	ExperimentID int
+	Score        float64
```
dpordomingo, line 34: What `Score` is? We didn't consider that in our initial schema...
Do you have any doc about it?

server/dbutil/dbutil.go
```diff
@@ -56,7 +55,7 @@ const (
 		FOREIGN KEY(experiment_id) REFERENCES experiments(id))`
 	createAssignments = `CREATE TABLE IF NOT EXISTS assignments (
 			user_id INTEGER, pair_id INTEGER, experiment_id INTEGER,
-			answer INTEGER, duration INTEGER,
+			answer TEXT, duration INTEGER,
 			PRIMARY KEY (user_id, pair_id),
```
dpordomingo: We considered that the same `FilePair` can be assigned to the same `User` many times in different experiments, so it can not be `PRIMARY KEY`

server/handler/user.go
```diff
@@ -13,13 +12,13 @@ import (
...
-			return nil, fmt.Errorf("no user id in context")
+		userID, err := service.GetUserID(r.Context())
+		if err != nil {
+			return nil, err
 		}
 
-		u, err := usersRepo.Get(uID)
-		if err != nil {
+		u, err := usersRepo.GetByID(userID)
+		if err != nil || u == nil {
```
dpordomingo: `404` only if `!error` and `!user`
```go
if err != nil {
    return nil, err
}
if user == nil {
    return nil, serializer.NewHTTPError(http.StatusNotFound, "user not found")
}
```

server/repository/assignments.go
```diff
@@ -0,0 +1,151 @@
...
+
+// Create stores an Assignment into the DB, and returns that new Assignment
+func (repo *Assignments) Create(as *model.Assignment) error {
+
+	_, err := repo.db.Exec(
+		`INSERT INTO assignments (user_id, pair_id, experiment_id, answer, duration)
+		VALUES ($1, $2, $3, $4, $5)`,
+		as.UserID, as.PairID, as.ExperimentID, as.Answer, as.Duration)
+
+	if err != nil {
```
dpordomingo: Can the insertion succeed, but the `Get` fail? And if so, does it mean that the creation failed? I don't think so...
At this point, I'd replace the if and everything below this comment with:
```go
return err
```

server/repository/assignments.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil, fmt.Errorf("Error getting assignment from the DB: %v", err)
+	default:
+		return &as, nil
+	}
+}
+
+// Get returns the Assignment for the given user and pair IDs. If the Assignment
+// does not exist, it returns nil, nil
+func (repo *Assignments) Get(userID, pairID int) (*model.Assignment, error) {
+	return repo.getWithQuery(repo.db.QueryRow(
```
dpordomingo: This can return many assignments:
>_We considered that the same `FilePair` can be assigned to the same `User` many times in different experiments, so it can not be `PRIMARY KEY`_

server/repository/assignments.go
```diff
@@ -0,0 +1,151 @@
...
+	if len(results) == 0 {
+		return nil, ErrNoAssignmentsInitialized
+	}
+
+	return results, nil
+}
+
+// Update updates the Assignment identified by the given user and pair IDs,
+// with the given answer and duration
+func (repo *Assignments) Update(userID int, pairID int, answer string, duration int) error {
```
dpordomingo: `s/pairId/assignmentId`

server/repository/assignments.go
```diff
@@ -0,0 +1,151 @@
...
+	if _, ok := model.Answers[answer]; !ok {
+		return fmt.Errorf("Wrong answer provided: '%s'", answer)
+	}
+
+	cmd := fmt.Sprintf(
+		"UPDATE assignments SET answer='%v', duration=%v WHERE user_id=%v AND pair_id=%v",
```
dpordomingo: >_We considered that the same `FilePair` can be assigned to the same `User` many times in different experiments, so it can not be `PRIMARY KEY`_

The idea of checking the `user_id` is redundant if it is used the `assignmentId`, but it can be a good idea to ensure that the assignment being saved, belongs to the user in session.

server/repository/experiments.go
```diff
@@ -0,0 +1,71 @@
...
+func (repo *Experiments) Create(exp *model.Experiment) error {
+	// TODO: for now this method is not used, but if we allow experiment creation
+	// the name should be safely escaped
+	return fmt.Errorf("Not implemented")
+
+	_, err := repo.db.Exec(
+		"INSERT INTO experiments (name, description) VALUES ($1, $2)",
+		exp.Name, exp.Description)
+
+	if err != nil {
```
dpordomingo: _idem_
Can the insertion succeed, but the `Get` fail?
```go
return err
```

server/repository/users.go
```diff
@@ -1,34 +1,65 @@
...
+}
+
+// Create stores a User into the DB, and returns that new User
+func (repo *Users) Create(user *model.User) error {
+
+	_, err := repo.db.Exec(
+		"INSERT INTO users (login, username, avatar_url, role) VALUES ($1, $2, $3, $4)",
+		user.Login, user.Username, user.AvatarURL, user.Role)
+
+	if err != nil {
```
dpordomingo: _idem_
Can the insertion succeed, but the `Get` fail?
```go
return err
```

server/repository/experiments.go
```diff
@@ -0,0 +1,71 @@
...
+	case err != nil:
+		return nil, fmt.Errorf("Error getting experiment from the DB: %v", err)
+	default:
+		return &exp, nil
+	}
+}
+
+// Get returns the Experiment with the given name. If the Experiment does not
+// exist, it returns nil, nil
+func (repo *Experiments) Get(name string) (*model.Experiment, error) {
```
dpordomingo: Do we assume this method is going to be needed in the future?
I can not imagine with which purpose... Any idea?

If it makes sense to need it, I'd rename by `GetByName` (same as `GetById`)

server/service/oauth.go
```diff
@@ -41,7 +39,8 @@ func NewOAuth(clientID, clientSecret string) *OAuth {
 	}
 }
 
-type githubUser struct {
+// GithubUser represents the user response returned by the GitHub auth.
+type GithubUser struct {
```
dpordomingo: I think it does not need to be public, because handlers does not build it, just use it

dpordomingo (changes requested):
Wow! that's great, I only found some things (repeated in some places), here is the list:
- one `FilePair` can be assigned to the same `User` many times in different experiments, so it can not be `PRIMARY KEY`
- avoid named returns
- use http status `500` when user can not be fetched
- what `FilePair.Score` is?
- the error returned by `repository.Whatever.Create(Whatever)` must be the the error returned by the `insert` query
- do we really need `repository.Experiment.GetByName(string)?

server/dbutil/dbutil.go
smacker: don't think currently it's a big problem. The problem is FE expects to have an id for assignment because it's standard in REST.

server/repository/experiments.go
smacker: I would propose to avoid `Not implemented` methods at all as long as we don't need them.

server/dbutil/dbutil.go
carlosms: According to [the schema](https://github.com/src-d/code-annotation/issues/2#issuecomment-357315069), a FilePair belongs to one and only one experiment.

server/handler/assignments.go
carlosms: This makes sense assuming the current primary key of the assignments table is correct.
Right now there isn't an assignment ID (again, following  [the schema](https://github.com/src-d/code-annotation/issues/2#issuecomment-357315069)).

If this looks bad, we could actually change the route to be `PUT /experiments/{experimentId}/`, and get hte pairID from the body.

server/model/models.go
carlosms: Score is the similarity score, it comes from the imported data.
It was added when real data was provided by the ML team, see [here](https://github.com/src-d/code-annotation/issues/2#issuecomment-359045024).

server/handler/assignments.go
smacker: please add `assignmentId ` to schema to follow rest conventions and don't rewrite frontend (I believe it's easier to add id, than change FE)

server/dbutil/dbutil.go
dpordomingo: mmm... I did't fond where it was written that a FilePair belongs to only one experiment. It's true that the oposite was not written too, but I feel it's more compatible with the user necessities.

server/repository/experiments.go
dpordomingo: I think that non implemented methods are a good way to define the architecture and the API, but only when it is really possible to need them. But in this case, I don't see when are we using `repo.Experiment.GetByName`

smacker: it's exactly what I mean. If we don't see where we are going to use them soon ("don't need" in my previous comment), let's avoid them.

server/dbutil/dbutil.go
carlosms: Well, in the schema the file pairs table has an experiment ID column, I understood that as a relationship with one and only one experiment.

But as we discussed via chat, both of you you agree that is a requirement, so I'll change it to be a * to * relationship in separate table, and perform a join when selecting all file pairs belonging to an experiment in Assignments.Initialize

bzz:
@carlosms to speed up the reviews, could you please summarize the current status of the PR? 

Asking, as In PR description there are some TODO items that are not checked and then if I'm not mistaken, there seems to be some un-addressed feedback. 

If addressing the reviews requires substantial time, I find very useful putting together a [comment with markdown checkboxes](https://github.com/src-d/code-annotation/pull/16#issuecomment-358992502), to visualize the progress better and this way to communicate the status with reviewers.

carlosms:
@bzz the description checkboxes were TODOs I found myself to remind me of doing that when the final version before the merge was ready.

- [X] Update README with new env variable (done in e23160c)
- [x] Check if DB strings need to be escaped

The current un-adressed feedback are the following two changes:

- [x] Add an assignment ID as the primary key of the assignments table
- [ ] ~~Allow file-pairs to belong to several experiments~~ New issue: https://github.com/src-d/code-annotation/issues/66

server/repository/users.go
carlosms: Even if it's unlikely, the Get may fail if the DB connection breaks, or someone deletes the entry at that very exact moment.
In that case the user pointer is not correctly populated with the assigned ID, and I'd say the method should notify of this returning an error.

carlosms:
Do you @smacker & @dpordomingo think we can leave the requirement 'Allow file-pairs to belong to several experiments' out of this PR?

I propose to open an issue to do it later it in a different PR once this branch is in master. Reasoning:
- It affects code that was already in master
- it's not really needed anytime soon, we actually only have one hardcoded experiment.
- this way we can merge this PR and #56 as soon as possible.

server/handler/assignments.go
```diff
@@ -45,14 +42,30 @@ type assignmentRequest struct {
...
+			return nil, serializer.NewHTTPError(http.StatusNotFound, "assignment not found")
+		}
+
+		userID, err := service.GetUserID(r.Context())
+		if err != nil {
+			return nil, err
+		}
+
+		if userID != assignment.UserID {
+			return nil, serializer.NewHTTPError(http.StatusUnauthorized,
```
smacker: 1. `StatusForbitten` may be better here. It's a common approach to `logout` user on `StatusUnauthorized` when there is a separate frontend application. Because you usually return this error in auth middleware. (for example when token is outdated)
2. is it better to `Get(id, userID)`?

not a blocker though

server/repository/file_pairs.go
```diff
@@ -0,0 +1,47 @@
...
+func NewFilePairs(db *sql.DB) *FilePairs {
+	return &FilePairs{db: db}
+}
+
+// getWithQuery builds a FilePair from the given sql QueryRow. If the FilePair
+// does not exist, it returns nil, nil
+func (repo *FilePairs) getWithQuery(queryRow *sql.Row) (*model.FilePair, error) {
+	var pair model.FilePair
+
+	var a, b model.File
```
smacker: something weird is going on here. You never use `a`&`b` beside scanning values to them.

server/repository/users.go
```diff
@@ -1,34 +1,65 @@
...
+		return err
 	}
-	r.users = append(r.users, m)
-	return nil
+
+	user, err = repo.Get(user.Login)
```
smacker: do you need this get to get `id`?
this way it's not going to work. Simplified example of what you are doing: https://play.golang.org/p/dlEmRgTL1uN

server/service/jwt.go
```diff
@@ -77,8 +77,19 @@ func (j *JWT) Middleware(next http.Handler) http.Handler {
...
 
-// GetUserID gets user id setted by JWT middleware
-func GetUserID(ctx context.Context) int {
-	id, _ := ctx.Value(userIDKey).(int)
-	return id
+// getUserInt gets the value stored in the Context for the key userIDKey, bool
+// is true on success
+func getUserInt(ctx context.Context) (int, bool) {
+	i, ok := ctx.Value(userIDKey).(int)
+	return i, ok
```
smacker: `return ctx.Value(userIDKey).(int)`

server/repository/file_pairs.go
carlosms: Good catch, this should be pair.Left and Right. I missed it because the serializer does nothing with these two values.

server/repository/users.go
dpordomingo: mmm... I see two patterns:
- `error := userRepo.Create(user)` -> it should not modify the user, but return an error if it failed storing in the DB
- `error := user.Save()` -> it stores the user, and assigns its ID

If we go with the first one, the ID should be assigned from the outside of the Create method, I mean: instead of hiding the `.Get` call inside the `.Create` method, doing it from the outside:
```go
if error := userRepo.Create(userToStore); err != nil { 
    return err
}
// in some cases, we won't need to continue, so the Get call won't be
// neccessary, but if we need a user with ID, we can continue:
user, err := userRepo.Get(user.Login)
if err != nil { 
    return err
}
//at this point, we have a valid user
fmt.Println(user)
```
^ would you approve that @smacker ?

I saw that many times to avoid hiding complexity in methods that seem to be "low level" (and `userRepo.Create` seems to be that kind of methods)

dpordomingo: *PS:*
if it is assumed a possible race condition and the necessity of calling `.Get` from the `.Create` to modify the `user` retrieved as an input parameter, it can happen unexpected things;
It can happen that the final (modified) `user` will be quite different than the input one (another Name, Login...) so the caller function will operate with that `user` without noticing it was "changed in behind" by the `.Create` func.

(that's another reason because it could be a bad idea to modify the input user inside the `.Create` func)

smacker (approved):

dpordomingo (approved):
LGTM; you did a great work @carlosms  
Many thanks!!!

I have only one concern about the final implementation as explained in the comment https://github.com/src-d/code-annotation/pull/52#discussion_r166358719 and https://github.com/src-d/code-annotation/pull/52#discussion_r166381084, but if @smacker disagrees I can stand without it.

The thinks you marked as "TODO", imo can be done in a separated PR, and they're not blocker:
- [ ] Update README with new env variable -> just agree with @smacker how to push to #29
- [ ] Allow file-pairs to belong to several experiments -> your current implementation seems to follow our specs, and it works; it could be done in the future if really needed.

Before merging I'd only require:
- [ ] open an issue explaining the (possible) problem with unescaped things,
- [ ] rebase and squash to obtain a history that makes sense to you.

server/repository/users.go
smacker: I'm okay with current solution and your proposal is okay too. Both approaches are common. For error in `get` and not in `exec` people invented transactions if we really need it.

I would suggest to merge it as is now.

dpordomingo:
Since @smacker is ok with both approaches https://github.com/src-d/code-annotation/pull/52#discussion_r166388220 lets merge when @bzz approves the PR and: 
- [ ] open an issue explaining the (possible) problem with unescaped things, if necessary,
- [ ] rebase and squash to obtain a history that makes sense to you.

bzz:
Both, handling `'Allow file-pairs to belong to several experiments' out of this PR` and adding usage of the transactions for complex failure modes in subsequent PRs makes sense to me.

Great job @carlosms !

Probably a nitpick but 2 things:

 - [x] One thing I find useful is to have string literal extracted to the constants inside the code.
   The reasons are:
   - it makes SQL queries more visible
   - they are grouped in the single place
   - it simplifies the subsequent refactoring 
   - and facilitates re-use (easier to notice if you already have it defined)

 - [x] Another minor but nice practice is
    In case you are using `//TODO` to label some short-term subsequent fixes is to save some time for a readers by adding `//TODO(gh-username)` so no need for `git glame` while looking at it later.

Would you mind adding those?

One very last and very minor thing, but for me, as a reviewer, it's a bit confusing to merge PR with todo checkboxes not checked in PR description/comments of PR author - they are communicating the idea of more work needed to be done. It would help to avoid the confusion if you could update https://github.com/src-d/code-annotation/pull/52#issue-293184299 and  https://github.com/src-d/code-annotation/pull/52#issuecomment-363378896 . 

Feel free to use ~strikethough~ for items that are intended not to be checked in this PR, as they have a link to elsewhere, where this item is been taken care of.

bzz (approved):
LGTM.

Thank you Carlos!

carlosms:
Replying @dpordomingo 
> open an issue explaining the (possible) problem with unescaped things, if necessary,

With the final code strings are inserted using the `$N` query placeholders, which should be safe. The only place where a "raw" string is used is [here](https://github.com/src-d/code-annotation/pull/52/files#diff-ebc27875bd59733c7382e0a9013ff999R131), but the string is validated to be one of `model.Answers`. So I won't open an issue, no work is needed.

In reply to @bzz 
> Both, handling 'Allow file-pairs to belong to several experiments' out of this PR

New issue created: https://github.com/src-d/code-annotation/issues/66

> and adding usage of the transactions for complex failure modes in subsequent PRs makes sense to me.

Are you refering to this comment https://github.com/src-d/code-annotation/pull/52#discussion_r166388220 ? I don't think a transaction is the proposed solution, as @smacker suggested to leave the current code as it is.

bzz:
Sounds good to me. Let's :shipit:

PR #56, opened 2018-02-01
Connect the frontend UI with the backend API

dpordomingo:
List of changes:
- c029aba Make Assignment.Answer nullable
- 901345b Fix push FilePair api url
- b603e55 Use backend instead of mocks in front
- 371c146 Implement the answering from UI

Other minor changes:
- f2a0535 Emit valid diffs
- 0755005 Use answers as defined by the Backend


src/api/index.js
```diff
@@ -48,6 +47,10 @@ function normalizeErrors(err) {
 function apiCall(url, options = {}) {
   const token = TokenService.get();
 
+  if (options.body) {
+    options.body = JSON.stringify(options.body)
```
smacker: content-type is missed

src/api/index.js
```diff
@@ -89,13 +101,8 @@ let exportObject = {
...
+  putAnswer,
 };
-if (process.env.NODE_ENV !== 'test') {
-  exportObject = {
-    ...mocks,
-    me,
-  };
-}
 
 // eslint-disable-next-line
```
smacker: no need for it anymore

src/state/experiment.js
```diff
@@ -125,6 +126,19 @@ const loadFilePairIfNeeded = id => (dispatch, getState) => {
   );
 };
 
+const pushAnswer = (experimentId, assignmentId, answer) => (dispatch, getState) => {
```
smacker: it's not push, it's put or update :)

src/state/experiment.js
```diff
@@ -125,6 +126,19 @@ const loadFilePairIfNeeded = id => (dispatch, getState) => {
...
+  const { experiment } = getState();
+  const assigment = experiment.assignments.find(a => a.id === assignmentId);
+  const answerPayload = {
+    answer: answer,
+    duration: assigment.duration
+  }
+  return api.putAnswer(experimentId, assignmentId, answerPayload).catch(e => {
+    dispatch(addErrors(e));
+    dispatch({ type: LOAD_ERROR, error: e });
```
smacker: it will be super weird behavior and also race condition is possible

smacker (changes requested):

smacker:
error handling is missed also. Take a look here: https://github.com/smacker/code-annotation/commit/7340215ba7e96494f264658784e84b793152e3eb#diff-284817f1b6aad47c0ac6d61bd72f474c

dpordomingo:
@smacker feedback addressed by 29121ae

about error handling from https://github.com/src-d/code-annotation/pull/56#issuecomment-362346622
I didn't want to do more things than required. I just added what it was needed to connect with the backend API regardless previous state (that didn't handle errors).
I think it should be better do it in a separated PR as a global thing, shouldn't be?

smacker:
In my opinion correct api handler is part of connection issue. You connect not only successful case but unsuccessful too. But if you prefer to do it in another PR it’s fine.

src/api/index.js
dpordomingo: ???

src/state/experiment.js
dpordomingo: I copied it from other places
https://github.com/src-d/code-annotation/blob/b2e1db20a8498b372e88f9ed58b0e378eb8b405e/src/state/experiment.js#L136
https://github.com/src-d/code-annotation/blob/b2e1db20a8498b372e88f9ed58b0e378eb8b405e/src/state/experiment.js#L179
Did I miss something? Any `dispatch` maybe?

src/api/index.js
smacker: I added it because there were mocks.
Now we can safely remove it.
Please take a look how I did it when connected API to UI:
https://github.com/smacker/code-annotation/commit/7340215ba7e96494f264658784e84b793152e3eb#diff-284817f1b6aad47c0ac6d61bd72f474c

src/state/experiment.js
smacker: Yes, you are missing the logic. Of messed up with how async promises work.
Let me explain the whole flow: 

A user opens page `/exp/1/1`, what happens next:
1. middleware see the change and call `selectAssigment` action
2. `selectAssigment` sets `assignmentId` and calls `loadFilePairIfNeeded` action
3. If there are no files in cache, it calls API and returns promise
4. When promise is resolved it sets file content to store or if promise is rejected it sets error
5. User choose are those files similar or different. When user clicks button it calls `mark` action.
6. `mark` action sets answer to store
> next your changes that work wrong
7. You call `pushAnswer` action, it returns promise and sets error
8. __at the same time__ (next tick) it calls `nextAssigment` which starts changes url and everything starts over.

It's very possible that next assignment will be loaded and shown to user __before__ `PUT` api call has finished. What would happen in such case?
1. User will see next assignment and start deciding answer for the next assignment
2. Then he will get error at very random time

right now we don't reset errors anywhere but if we do later it will be even worst.
Because you don't wait for the promise, another promise can remove your error.
It will look like this for user:
1. Click button
2. Next assignment loader
3. Some weird flash
4. Next assignment, but the previous answer wasn't saved on server

bzz:
@dpordomingo could you please help me to understand, what is the status of this PR?

There seems to be a CI failure and some feedback may still need to be addressed.

dpordomingo:
@bzz CI success, but this is still pending https://github.com/src-d/code-annotation/pull/56#discussion_r165810881.

@smacker do you think it should be fixed before merging?
If possible, I'd merge and continue working in that (that way, @carlosms could rebase its work from https://github.com/src-d/code-annotation/pull/52 on top of this one, and test its implementation with a real working site)
If you think we can merge without https://github.com/src-d/code-annotation/pull/56#discussion_r165810881 tell me, and I'll create an issue, rebase, merge, and continue working in that request.

smacker:
@dpordomingo current code doesn't work correctly. It's not just missed feature it's a bug.

To fix it just change:

```javascript
dispatch(pushAnswer(experimentId, assignmentId, answer));
return dispatch(nextAssigment());
```

```javascript
return dispatch(pushAnswer(experimentId, assignmentId, answer)).then(() => {
  return dispatch(nextAssigment());
});
```

there can be a smarter solution, but this works and good enough to allow PR get merged.

dpordomingo:
Many thanks @smacker for the hint you offered by https://github.com/src-d/code-annotation/pull/56#issuecomment-363788395 ; I added it here: d2522bd

I also had to fix things after rebasing agains current `master`:
- c16121f Add default internal db to the .gitignore as agreed with @carlosms 
- 737db98 Implement Scanner and Valuer for model.Role as explained irl to @carlosms and @smacker 
- 71b2e65 Return 404 error when no assignments available to be labeled

server/model/models.go
```diff
@@ -51,6 +56,34 @@ type File struct {
 // Role represents the position of a app User
 type Role string
 
+// Value returns the string value of the Role
+func (r Role) Value() (driver.Value, error) {
+	// TODO: role validation (@dpordomingo)
```
bzz: Usual way of formatting todos is

`//TODO(dpordomingo): role validation`

server/model/models.go
```diff
@@ -51,6 +56,34 @@ type File struct {
...
+func (r *Role) Scan(value interface{}) error {
+	// TODO: role validation (@dpordomingo)
+	if value == nil {
+		*r = Role(Worker)
+		return nil
+	}
+
+	var bv driver.Value
+	var err error
+	if bv, err = driver.String.ConvertValue(value); err == nil {
```
smacker: why so complicated? https://play.golang.org/p/3ynyW8ZGKJ5 covers everything we need, no?

server/handler/assignments.go
```diff
@@ -28,6 +28,10 @@ func GetAssignmentsForUserExperiment(repo *repository.Assignments) RequestProces
 		assignments, err := repo.GetAll(userID, experimentID)
 		if err == repository.ErrNoAssignmentsInitialized {
 			if assignments, err = repo.Initialize(userID, experimentID); err != nil {
+				if err == repository.ErrNoPairsAvailable {
```
smacker: it's incorrect if we are following REST. List APIs should return an empty list when nothing is found, but a request is correct. There is no special rule for it because actually spec for REST doesn't exist. But if you look at any REST API (github for example) you will notice it works like that.

src/api/index.js
```diff
@@ -79,23 +82,24 @@ function getAssignments(experimentId) {
...
+  return apiCall(
+    `/api/experiments/${experimentId}/assignments/${assignmentId}`,
+    {
+      method: 'PUT',
+      body: answer,
+    }
+  );
 }
 
 // eslint-disable-next-line
```
smacker: as I mentioned before we don't need to disable eslint anymore.

smacker (changes requested):
please don't break common REST conventions.

server/handler/assignments.go
smacker: 404 error for endpoint `/api/:id/list` means there is no such API or id isn't found. Not empty list.

dpordomingo: `/api/experiments/{experimentId}/assignments` is requesting for the assignments of a given experiment, so this `404 Not Found` means that the collection does not exist at all. It is not an "empty" case.

About the status code: `404 Not Found` it's defined as 
>_The requested resource could not be found but may be available in the future. Subsequent requests by the client are permissible._

what matches our case: if in the future are loaded file-pairs, the endpoint will return data.

dpordomingo: `/api/experiments/{experimentId}/assignments` is requesting for the assignments of a given experiment, so this `404 Not Found` means that the collection does not exist at all. It is not an "empty" case.

About the status code: `404 Not Found` it's defined as 
>_The requested resource could not be found but may be available in the future. Subsequent requests by the client are permissible._

what matches our case: if in the future are loaded file-pairs, the endpoint will return data.

dpordomingo:
@bzz @smacker Your suggestions were addressed by 2623b4fd26d3437c58a9b0baa83d9b6e6a1e0a4e
I didn't change the response to an unexistent collection of assignments because since the collection does not exist, it makes sense the `404` (it is not an empty collection, but a collection that does not exist)
I think that further discussions should be addressed aside.

dpordomingo:
I don't think it worths that amount of discussions for something already done, that is consistent with REST and HTTP status codes and fits our necessities.

Anyhow we need to go prod as soon as possible and it's better to move on, so I dropped the `ErrNoPairsAvailable` error https://github.com/src-d/code-annotation/pull/56/commits/65c4e0a2c3b8015974d030f19516c52b17e775a7.
The amount of code removed is not relevant at all, and now the frontend also needs changes, but it can be done in a separated PR:

When the user has no assignments to fill, instead of rising a reasonable error, he obtains a "success for filling `NaN% experiments`, with no possibility to move to any other window.

image
issue: https://github.com/src-d/code-annotation/issues/82

smacker (approved):
Looks great now! Let's merge!!!!

bzz (approved):
LGTM

Awesome ✨ !

dpordomingo:
:cry: I wanted to rebase and reorder de history...

Issue #47, opened 2018-01-30
Reviewer: new page on the frontend, showing ML-extracted features


bzz:
Implement a page, with the reviewer part of annotation tool.

This would require proposing a JSON format for backend, to be used in #44 and mocking the KV results, until backend endpoint  is in place.

 - [x] Add new page to the FE #57 
-  [x] Endpoint with results of annotations #88
 - [x] make this page use backend from #44
 - [x] add restrictions to backend endpoints & frontend router #124

~This would also require something like hard-coding a "reviewer" bit for some users inside the DB, as proper user and ACL is outside of the scope right now.~

ACL is handled in #60

smacker:
the initial frontend is done. I believe for API side we have another issue, but just in case I'm moving this one back in to do.

dpordomingo:
@smacker could you create an issue (or put it here) the specs of the new endpoints that will be needed to achieve this feature?

bzz:
@dpordomingo AFAIK new endpoint is handled under #44 .

@smacker Initial work on FE is done in #57, so please, correct me if I'm wrong, but what's left under this issue is to make the page display the results from the API.

PR description updated.

smacker:
yes. we all set for api with #44. Just need to make sure our existing API returns blob id and score from database.

smacker:
done  🎉 further improvements will be handled in other issues

bzz:
🎉 For the history, could you please also post the URL of the reviewer page here?

smacker:
`http(s)://<UI-HOST>/review`

Issue #44, opened 2018-01-30
Reviewer: new backend endpoint, serving extracted features


bzz:
To show the extracted features in Reviewer UI, we would need a new handler on a backend that:
 - given 2 files
 - reads it's features from a new table in DB (table creation using a one-off script is handled elsewhere)
 - exposes them to the FM in KV format

For the context: reading extracted features from DB is a temporary solution. In subsequent PRs (after Feb 10th), this handler going to be replaced with a call though the REST API to a ML tool, that going to compute those feature on the fly, given the 2 files.

PR #57, opened 2018-02-01
Reviewer UI (first version)

smacker:
Part of https://github.com/src-d/code-annotation/issues/47

Based on https://github.com/src-d/code-annotation/pull/45

It's a little different from wireframes:
1. There are no breadcrumbs
2. Tables aren't scrollable, but tab content is. It's huge pain to make TBODY scrollable (actually you can't without overriding table layout by css) And then they are inside another flexible div. It's nightmare. I spent half of the day and almost implemented it. But it has too many hacks and I found few cases when it works wrong. We definitely need separate task for it or maybe @ricardobaeta can come up with another solution
3. Better select control is implemented in @ricardobaeta PR, we will replace it later

P.S. There is some magic of mixing grid with flexbox because I had no choice.

screen shot 2018-02-02 at 14 39 28

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "react-dom": "^16.2.0",
     "react-redux": "^5.0.6",
     "react-scripts": "1.0.17",
+    "react-split-pane": "^0.1.74",
     "redux": "^3.7.2",
     "redux-devtools-extension": "^2.13.2",
     "redux-little-router": "^14.2.3",
--- a/src/App.js
+++ b/src/App.js
@@ -5,6 +5,7 @@ import Errors from './components/Errors';
 import Index from './pages/Index';
 import Experiment from './pages/Experiment';
 import Final from './pages/Final';
+import Review from './pages/Review';
 
 class App extends Component {
   render() {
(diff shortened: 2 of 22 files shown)


bzz:
@smacker great! 

For changes that touch UI part, I find that also having a screenshot attached to PR helps to speed up the review process.

smacker:
@bzz screenshot added

bzz (approved):
Looks great to me.

I guess there will be subsequent tasks on using a mock data in components to show features, and then replacing it with real data from endpoint in #44

carlosms:
Looks like a very good first version

I've seen just one issue, the contents appear to go past the width no matter my window size:
127 0 0 1_8080_review

And a couple of comments:
- I'd prefer to see the mock data somehow separated, and the table code close to final
- If we merge this as a first version, let's add a comment to https://github.com/src-d/code-annotation/issues/47 saying that breadcrumbs and scrollable table are missing

carlosms (approved):

smacker:
Thanks, @carlosms I didn't have the bug with the width of the table. Will try to reproduce.

I want to rebase this PR using new components, so @ricardobaeta will be able to start doing visual design on this page without reimplementing the work he did already for annotation page.

smacker:
Rebased on #65 to unlock Ricardo.
It looks like this now:

screen shot 2018-02-08 at 12 58 09

Problem with table width is fixed.

Issue #48, opened 2018-01-30
Restrict access of authenticated users


bzz:
To deploy an app, we want to be able to restrict access to annotation, limiting it to just some Gitub users.

Same for reviewers - even smaller set of users would need a reviewer bit set on.

Developing a full-blown ACL solution is outside of the scope though.

PR #60, opened 2018-02-06
Restrict access of authenticated users

smacker:
Fixes: https://github.com/src-d/code-annotation/issues/48


bzz (approved):
LGTM

\cc @carlosms for review

server/service/oauth.go
```diff
@@ -15,29 +17,37 @@ import (
 	"golang.org/x/oauth2/github"
 )
 
+var NoAccessError = errors.New("access deniend")
```
carlosms: Exported without comment

server/service/oauth.go
```diff
@@ -88,12 +98,83 @@ func (o *OAuth) GetUser(ctx context.Context, code string) (*model.User, error) {
...
 		Username:  user.Username,
 		AvatarURL: user.AvatarURL,
+		Role:      role,
 	}, nil
 }
+
+const orgPrefix = "org:"
+const teamPrefix = "team:"
+
+func (o *OAuth) checkAccess(client *http.Client, restriction, login string) error {
```
dpordomingo, line 135: does not need access to `(o *OAuth)`, does it?

server/service/oauth.go
```diff
@@ -88,12 +98,83 @@ func (o *OAuth) GetUser(ctx context.Context, code string) (*model.User, error) {
...
+	role := model.Requester
+	if o.restrictReviewerAccess != "" {
+		err := o.checkAccess(client, o.restrictReviewerAccess, user.Login)
+		if err != nil && err != NoAccessError {
+			return nil, err
+		}
+		if err == NoAccessError {
+			role = model.Worker
+		}
+	}
```
carlosms: With the code above, the reviewers are a subgroup of the workers, right?

This is not wrong, but it does not match with what we discussed IRL. We said that for greater flexibility we would allow independent orgs/teams for each role.

What about something like this?
```
	role := model.Requester

	if o.restrictReviewerAccess != "" {
		err := o.checkAccess(client, o.restrictReviewerAccess, user.Login)
		if err != nil && err != NoAccessError {
			return nil, err
		}
		if err == NoAccessError {
			role = model.Worker
		}
	}

	if role == model.Worker && o.restrictAccess != "" {
		if err := o.checkAccess(client, o.restrictAccess, user.Login); err != nil {
			return nil, err
		}
	}
```

server/service/oauth.go
```diff
@@ -88,12 +98,83 @@ func (o *OAuth) GetUser(ctx context.Context, code string) (*model.User, error) {
...
+		return nil, fmt.Errorf("can't parse github user response: %s", err)
+	}
+
+	if o.restrictAccess != "" {
+		if err := o.checkAccess(client, o.restrictAccess, user.Login); err != nil {
+			return nil, err
+		}
 	}
+
+	role := model.Requester
```
carlosms: We are using the `reviewer` and `requester` words for the same concept.
Maybe this PR is a good place to fix this.

server/service/oauth.go
```diff
@@ -88,12 +98,83 @@ func (o *OAuth) GetUser(ctx context.Context, code string) (*model.User, error) {
...
+			return fmt.Errorf("can't get user organizations from github: %s", err)
+		}
+		// StatusNoContent means user is a member
+		// https://developer.github.com/v3/orgs/members/#check-membership
+		if resp.StatusCode != http.StatusNoContent {
+			return NoAccessError
+		}
+		return nil
+	}
+	if strings.HasPrefix(restriction, teamPrefix) {
```
dpordomingo: we (at src-d) are separating code blocks with a new line in between.
([examples in gogit](https://github.com/src-d/go-git/blob/master/remote.go))

server/service/oauth.go
```diff
@@ -88,12 +98,83 @@ func (o *OAuth) GetUser(ctx context.Context, code string) (*model.User, error) {
...
+		resp, err := client.Do(r)
+		if err != nil {
+			return fmt.Errorf("can't get user organizations from github: %s", err)
+		}
+		// Only StatusOK means user is a member
+		// https://developer.github.com/v3/orgs/teams/#get-team-membership
+		if resp.StatusCode != http.StatusOK {
+			return NoAccessError
+		}
+		// Don't allow access to pending members
```
dpordomingo: Why do we need to forbid "pending" members? If we don't care about that, all code below this comment can be replaced by
```go
return nil
```

server/service/oauth.go
```diff
@@ -88,12 +98,83 @@ func (o *OAuth) GetUser(ctx context.Context, code string) (*model.User, error) {
...
+		}
+		if err = json.NewDecoder(resp.Body).Decode(&teamResp); err != nil {
+			return fmt.Errorf("can't parse github team response: %s", err)
+		}
+		if teamResp.State != "active" {
+			return NoAccessError
+		}
+		return nil
+	}
+	return errors.New("invalid restriction")
```
carlosms: Since this is an error intended for the admin responsible for the settings, and not a developer, I'd add more details to help:
`return errors.New("invalid restriction %s, it must be one of 'org:<organization-name>' or '<team:team-id>'") `

server/service/oauth.go
```diff
@@ -88,12 +98,83 @@ func (o *OAuth) GetUser(ctx context.Context, code string) (*model.User, error) {
...
+		Role:      role,
 	}, nil
 }
+
+const orgPrefix = "org:"
+const teamPrefix = "team:"
+
+func (o *OAuth) checkAccess(client *http.Client, restriction, login string) error {
+	if strings.HasPrefix(restriction, orgPrefix) {
+		org := strings.TrimPrefix(restriction, orgPrefix)
```
dpordomingo, line 137: too long.
we could extract the body of both ifs to separated functions.
```go
func (o *OAuth) checkAccess(client *http.Client, restriction, login string) error {
  if strings.HasPrefix(restriction, orgPrefix) {
    org := strings.TrimPrefix(restriction, orgPrefix)
    return checkUserInOrganization(client, org, login)
  }

  if strings.HasPrefix(restriction, teamPrefix) {
    team := strings.TrimPrefix(restriction, teamPrefix)
    return checkUserInTeam(client, team, login)
  }

  return return errors.New("invalid restriction")
}

func checkUserInOrganization(client *http.Client, org string, login string) error {}

func checkUserInTeam(client *http.Client, ream string, login string) error {}
```

README.md
```diff
@@ -27,6 +27,20 @@ cd $GOPATH/github.com/src-d/code-annotation
 make serve
 ```
 
+### Restrict access
+
+It's possible to restrict access to the tool or `reviewer` role by setting environmental variables:
```
dpordomingo: Why not add them to the `.env.tpl` file?

dpordomingo (changes requested):
I like your implementation!!!

I'd only require some things:
- `checkAccess` does not need to be an OAuth member, https://github.com/src-d/code-annotation/pull/60#discussion_r166583210
- `checkAccess` has a too long body that could be extracted in separate functions, https://github.com/src-d/code-annotation/pull/60#discussion_r166587761
- coding style https://github.com/src-d/code-annotation/pull/60#discussion_r166585556

questions:
- do we really need to exclude `pending` users? https://github.com/src-d/code-annotation/pull/60#discussion_r166587179

README.md
```diff
@@ -27,6 +27,20 @@ cd $GOPATH/github.com/src-d/code-annotation
...
+* `OAUTH_RESTRICT_ACCESS`
+* `OAUTH_RESTRICT_REVIEWER_ACCESS`
+
+As a value it accepts github organization name or github team id. Example:
+
+```bash
+OAUTH_RESTRICT_ACCESS=org:my-organization
+OAUTH_RESTRICT_REVIEWER_ACCESS=team:123456
+```
+
```
carlosms: May I suggest something more extended? The following text assumes that Reviewers are not a subgroup of the Workers. It is also worth noting that the current readme does not explain the two different roles.

### Access Control

It is possible to restrict access and choose each user's role by adding their GitHub accounts to a specific [organization](https://help.github.com/articles/collaborating-with-groups-in-organizations/) or [team](https://help.github.com/articles/organizing-members-into-teams/).

This is optional, if you don't set any restrictions all users with a valid GitHub account will be able to login as a Reviewer. You may also set a restriction only for Reviewer users, and leave open access to anyone as Workers.

To do so, set the following variables in you `.env` file:

* `OAUTH_RESTRICT_ACCESS`
* `OAUTH_RESTRICT_REVIEWER_ACCESS`

Both variables accept a string with either `org:<organization-name>` or `team:<team-id>`. For example:

```bash
OAUTH_RESTRICT_ACCESS=org:my-organization
OAUTH_RESTRICT_REVIEWER_ACCESS=team:123456
```

carlosms (changes requested):
Besides the inline comments, there is a big obstacle in my opinion:

Right now we ask the administrator to set a team-id, but it seems that you can't get that ID from the GitHub web interface.

We should do something about it, ideally the code should be able to work with a team name instead of the ID.
I think that using something like `team:<org-name>/<team-name>` we should be able to use this api call to the the team ID:
https://developer.github.com/v3/orgs/teams/#list-teams

server/service/oauth.go
smacker: doesn't, but I prefer to group functions as struct methods.

server/service/oauth.go
smacker: doesn't feel too long for me, but ok.

server/service/oauth.go
smacker: if we do that, we need linter for it. And please describe which blocks should be separated. From reading the source code of go-git I don't always see the pattern.

server/service/oauth.go
smacker: because pending member isn't member

smacker:
Comments are addressed. About team-id:
- Current API for teams isn't stable
- Get id is as simple as http://bfy.tw/GTDG

bzz:
Probably a minor thing, but CI seems to be failing, @smacker could you briefly summarize the reason and how is it taken care of?

All reviews seems to be addressed, so @carlosms @dpordomingo could you please make another round?

smacker:
@bzz restart of CI helped. It didn't clean up after itself first. (not sure why)

carlosms:
@smacker:

> Comments are addressed. About team-id:
> - Current API for teams isn't stable
> - Get id is as simple as http://bfy.tw/GTDG

As amusing your answer may be, the point still stands. We ask our users to provide the team-id without explaining what it is and how to obtain it.

Could you please try to come up with a more constructive proposal?
Or, if there isn't a clean way to manage teams, maybe we should consider it a limitation and remove it, allowing to filter only by organization membership.

smacker:
@carlosms haha. Let me rephrase. I propose to keep it as an internal feature. From requirements, I understood we __need__ teams, not only organizations. And for internal usage, we can grab ID easily. Maybe we need to just remove documentation about it.

The main reason why I don't want to write more code around it and expose it to external users:

> Warning: The API may change without advance notice during the preview period. Preview features are not supported for production use. If you experience any issues, contact GitHub support.
https://developer.github.com/v3/orgs/teams/

carlosms:
@smacker 

> I propose to keep it as an internal feature. From requirements, I understood we need teams, not only organizations. And for internal usage, we can grab ID easily. Maybe we need to just remove documentation about it.

I personally don't like the idea of having undocumented functionality, we either support teams or not. And if it's needed for our internal needs, that's a sign that it may be needed by other users.
But I think this decision should be probably made by @bzz

> The main reason why I don't want to write more code around it and expose it to external users:
> 
> Warning: The API may change without advance notice during the preview period. Preview features are not supported for production use. If you experience any issues, contact GitHub support.
> https://developer.github.com/v3/orgs/teams/

Then maybe we should be using the latest v4 API:
https://developer.github.com/v4/object/team/

smacker:
v4 is GraphQL API, not a REST one.

bzz:
@carlosms thank you for keeping the discussion focused and appreciate the constructive feedback on why teams are not exposed yet. I think it should be OK for now to keep it as is, merge PR and then add nice-to-have integration that uses the name/more docs around it.

@carlosms @dpordomingo could you please make another pass over it as the rest of the feedback seems to be addressed?

carlosms (approved):
Approved as requested by @bzz, but please open a new issue (or add a TODO comment in #48) to continue the conversation about team IDs.

smacker:
@carlosms the issue https://github.com/src-d/code-annotation/issues/72
I didn't change README because it's not a priority now.

server/service/oauth.go
dpordomingo: but they was added by the org's owner, what suits our purpose...

server/service/oauth.go
dpordomingo: Wow, that's a good catch: we decide to call that role "requester", but this PR introduces a new concept colliding with an existing one.
Let's do `s/reviewer/requester/g`

server/service/oauth.go
dpordomingo: wow... I think that's an ugly thing (and also forces extra steps in unit tests...) :(

server/service/oauth.go
```diff
@@ -78,17 +93,107 @@ func (o *OAuth) GetUser(ctx context.Context, code string) (*githubUser, error) {
...
 	defer resp.Body.Close()
+
 	var user githubUser
 	err = json.NewDecoder(resp.Body).Decode(&user)
 	if err != nil {
-		return nil, fmt.Errorf("can't parse github response: %s", err)
+		return nil, fmt.Errorf("can't parse github user response: %s", err)
 	}
 
+	role := model.Requester
```
dpordomingo, line 110: I think this code would be easier to read if it was:
```go
if o.restrictRequesterAccess != "" {
  if err := o.checkAccess(client, o.restrictRequesterAccess, user.Login); err == nil {
    user.Role = model.Requester
    return &user, nil
  } else if err != ErrNoAccess {
    return nil, err
  }
}

if o.restrictAccess != "" {
  if err := o.checkAccess(client, o.restrictAccess, user.Login); err == nil {
    user.Role = model.Worker
    return &user, nil
  } else {
    return nil, err
  }
}

user.Role = model.Requester
return &user, nil
```

If I'm not wrong, that way it's easier to understand the logic behind the role assignment depending on GitHub groups and the default value if nothing happens.

dpordomingo (commented):
This comments could be addressed by https://github.com/src-d/code-annotation/issues/72, so would not block the merge

.env.tpl
```diff
@@ -2,3 +2,5 @@ OAUTH_CLIENT_ID=
 OAUTH_CLIENT_SECRET=
 JWT_SIGNING_KEY=testing
 DB_CONNECTION=sqlite:///path/to/db.db
+OAUTH_RESTRICT_ACCESS=
+OAUTH_RESTRICT_REVIEWER_ACCESS=
```
dpordomingo: `Reviewer` instead of `Requesters`?
I don't see the reason because this PR introduces a new concept colliding with an already existing one. We decide to call that role `requester`, so it would be a good idea to stick to our convention,
I'd suggest doing `s/reviewer/requester/g` before merging.

dpordomingo (changes requested):
Everything is fine but one thing easy to change.
Once it is addressed, feel free to merge :smile:

server/service/oauth.go
dpordomingo: Most of our code is always written that way (with NewLine after every code block).
(random files, also in borges... https://github.com/src-d/borges/blob/master/archiver.go https://github.com/src-d/borges/blob/master/common.go)

(we would also need a linter to ensure the opposite; the reason is it's an internal convention we decided to follow...)

The rule could be:
>_a NewLine after `}` if it's not followed by another `}`_

.env.tpl
bzz: 👍 for consistency.

Let's rename and merge, with subsequent improvements planned in #72

smacker: let's try to use correct names everywhere. In the issue description, it's called `reviewers`.

smacker:
pushed rename reviewer -> requester.

.env.tpl
dpordomingo: It might be a misunderstanding when the latest issue was created.
In the [schema definition issue](https://github.com/src-d/code-annotation/issues/2) and in our code, we're already using `requester`.

dpordomingo (approved):
LGTM, many thanks!

Issue #64, opened 2018-02-07
Create reusable components on top of #53


smacker:
TBD

PR #53, opened 2018-02-01
In-Browser Visual Design

ricardobaeta:
@bzz @smacker @dpordomingo @carlosms 

This is the initial work on Code Annotation Visual Design. Only the annotation UI is designed yet. The other UI's are still subject to further design.

I would like to propose to apply a dark theme to the code visualisation and syntax highlighting as well.

Excuse me for the less mistakes, and some other changes that will imply changing code structure.

I have ideas on further developments that I will address as features requests.  

image

PR #65, opened 2018-02-07
Create components based on visual design

smacker:
Fixes: https://github.com/src-d/code-annotation/issues/64

This PR beings visual design from https://github.com/src-d/code-annotation/pull/53 but code is separated to reusable components.
It doesn't break other pages.

screen shot 2018-02-07 at 12 49 28

### DIFF
--- a/src/components/Breadcrumbs.js
+++ b/src/components/Breadcrumbs.js
@@ -1,0 +1,16 @@
+import React from 'react';
+import './Breadcrumbs.less';
+
+function Breadcrumbs({ items }) {
+  return (
+    <ul className="breadcrumbs">
+      {items.map((item, i) => (
+        <li key={i}>
+          <a href={item.link}>{item.name}</a>
+        </li>
+      ))}
+    </ul>
+  );
+}
+
+export default Breadcrumbs;
--- a/src/components/Breadcrumbs.less
+++ b/src/components/Breadcrumbs.less
@@ -10,6 +10,11 @@
 
     a {
         color: #818a8a;
+
+        &:hover {
+            text-decoration: none;
+            color: #fff;
+        }
     }
 
     > li {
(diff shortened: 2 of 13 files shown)


bzz:
#53 is merged now, feel free to rebase

carlosms (changes requested):
It looks great!

I don't know if it's because maybe there is some work @ricardobaeta did after you created this branch, but the header is different to master.
In this screenshot master is on top, this branch on the bottom:
screenshot from 2018-02-08 16-21-10

The differences are:
- Different left&right margins
- Missing breadcrumb items
- Wrong style for the dashboard link (it's not uppercase, and the hover color is not right)

Plus the breadcrumb links should probably have a different style on hover. Right now it's the only link with an underline on hover, and no color change. Maybe @ricardobaeta has a proposal for this.

smacker:
1. Margins come from bootstrap grid now. We can change them if we need. (it means we need to change grid)
2. The component itself is correct. The second part of breadcrumbs comes from a description. And it's empty. If there is anything like "asd" it will look the same. You can see it in https://github.com/src-d/code-annotation/issues/57 where I mocked all the data
3. This one is a mistake! Thanks for pointing to it!!! I'm fixing.

smacker:
links are fixed. All of them have the same hover style & uppercase.

ricardobaeta:
> Margins come from bootstrap grid now. We can change them if we need. (it means we need to change grid)

@smacker Let's start to tighten up our design. Please be so kind to address the margins issue.

carlosms (approved):
:+1:

ricardobaeta:
@smacker As per our conversation, it's all good with the margins. Thank you.

bzz (approved):
LGTM, thank you for clarifications!

PR #71, opened 2018-02-07
Server should bootstrap the DB if it's empty

### LINKED ISSUE #69: Server should bootstrap the DB if it's empty
Right now the server does not start if the DB is empty. It should be able to bootstrap the DB, and the import command should be able to add new data at some point later.

carlosms:
Fixes #69

### DIFF
--- a/cli/server/server.go
+++ b/cli/server/server.go
@@ -27,11 +27,19 @@ func main() {
 	logger := service.NewLogger()
 
 	// database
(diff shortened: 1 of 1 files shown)


smacker (approved):

bzz (approved):
LGTM

Thank you @carlosms for small PR and a prompt fix!

PR #76, opened 2018-02-08
Implement features endpoint

smacker:
Fixes https://github.com/src-d/code-annotation/issues/44

### DIFF
--- a/server/handler/features.go
+++ b/server/handler/features.go
@@ -1,0 +1,26 @@
+package handler
+
+import (
+	"net/http"
+
+	"github.com/go-chi/chi"
+	"github.com/src-d/code-annotation/server/repository"
+	"github.com/src-d/code-annotation/server/serializer"
+)
+
+// GetFeatures returns a function that returns a *serializer.Response
+// with the list of features for blobId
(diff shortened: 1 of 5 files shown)


bzz:
\cc @carlosms for review

carlosms:
Maybe we should restrict this route only for users with the Requester role. Do you agree?

smacker:
we don't have ACL yet. But we should later, yes.

carlosms:
The roles are already in models.go, and #60 should be merged soon. I'd make this change to save us a new PR in a few hours. Up to you.

Otherwise, looks good.

carlosms (approved):

smacker:
You won't be able to test it after merge then. I prefer going step by step. In my opinion, the right place to implement it is connection FE to BE. Because the same restriction will be needed on FE side.

carlosms:
I'm ok with any option as long as we don't forget about it. Can you please make a new issue, or add a comment to the relevant issue to make sure we don't forget? Thank you.

smacker:
@carlosms added in https://github.com/src-d/code-annotation/issues/47

bzz (approved):
LGTM

Issue #73, opened 2018-02-08
Implement a way to download DB exports


carlosms:
This issue is to implement a temporary (and very basic) way to use the web interface to:
- export the data to a SQLite DB
- download that file

bzz:
@carlosms do you think 

1) It may be worth documenting existence of the `/exports` somewhere as well, with link to schema, etc?

   So the users would be able to learn about it without deep-dive into the app code?

2) Also, as un-authorized user, I was able to access `/exports` page


<img width="1282" alt="screen shot 2018-02-14 at 4 59 58 pm" src="https://user-images.githubusercontent.com/5582506/36214123-e6f3b0bc-11a8-11e8-98a1-90c75ad46e4a.png">

dpordomingo:
just to clarify it... 2 is not a security problem but a UX thing (because no Requester can do nothing there)

ricardobaeta:
I'm here to help.

carlosms:
@bzz said:

> @carlosms do you think
> 
> It may be worth documenting existence of the /exports somewhere as well, with link to schema, etc?
> So the users would be able to learn about it without deep-dive into the app code?

Yes, I'll add it to the readme. But this route should probably have a link from somewhere in the app.
Can you please clarify what do you with link to schema? Do you want to have a new file with the schema for the `assignments` table and link to it from the readme?

> Also, as un-authorized user, I was able to access /exports page

As @dpordomingo said, that is not a security problem, you can access the route but you won't be able to do or see anything. Improving this is part of #47, and I guess we'll show an empty page with a message to let you know you are not authorized.

PR #81, opened 2018-02-09
DB Exports in the UI

carlosms:
Implements #73.

This PR adds handlers to allow the following actions from the UI:
- export the data to a SQLite DB
- download that file

The creation and download is restricted to users with the Requester role.
To access this new route, you must visit `<hostname>/exports`

_Note_: Of course there are many things to improve. The code has been done with the intention of being a quick and basic first iteration, for a functionality that we may not want to keep.
The code is all contained in a single file for this very same reason.

image

### DIFF
--- a/server/handler/exports.go
+++ b/server/handler/exports.go
@@ -3,7 +3,6 @@ package handler
 import (
 	"errors"
 	"fmt"
-	"html/template"
 	"io"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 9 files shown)


smacker:
Instead of comments I decided to show it in code. Here is PR https://github.com/carlosms/code-annotation/pull/1 (it also has commits from master, please rebase your branch then they will disappear)

The commit with improvements is here:
https://github.com/carlosms/code-annotation/pull/1/commits/372337c10c44c3ed55d755c933ec7854c3460c66

I just structured it to be a better fit for our application. Not much changes.

bzz:
structure of carlosms/code-annotation@372337c commit looks really nice, as example of adding a new feature \w backend and frontend.

What do you think, @carlosms ?

carlosms:
Maxim's commit is now in this branch. Please add your review @bzz @smacker

smacker (approved):
because now it includes my code I don't have comments 😆

bzz (approved):
Looks great to me.

As Ricardo says -  ✨ team work :)

bzz:
@carlosms do you think it may be worth documenting existence of the `/exports` somewhere as well, with link to schema, etc? 

So the users would be able to learn about it without deep-dive into the app code?

server/handler/exports.go
```diff
@@ -0,0 +1,111 @@
...
+		return nil, err
+	}
+
+	if err := dbutil.Copy(*h.db, destDB, dbutil.Options{}); err != nil {
+		return nil, err
+	}
+
+	service.NewLogger().Info("new SQLite file created: " + filepath)
+
+	return &serializer.Response{
```
dpordomingo, line 83: I think it would be better to avoid creating responses from scratch.
It could be somehow reused [serializer.newResponse(interface{}) *serializer.Response](https://github.com/src-d/code-annotation/blob/HEAD/server/serializer/serializers.go#L52)

server/handler/exports.go
```diff
@@ -0,0 +1,111 @@
...
+		Status: http.StatusOK,
+		Data:   filepath,
+	}, nil
+}
+
+// Download writes requested file to response
+func (h *Export) Download(w http.ResponseWriter, r *http.Request) {
```
dpordomingo, line 90: I think it could be somehow followed the design given by the static handler https://github.com/src-d/code-annotation/blob/HEAD/server/handler/static_files.go

server/service/acl.go
```diff
@@ -0,0 +1,42 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		userID, _ := GetUserID(r.Context())
+
+		user, err := s.usersRepo.GetByID(userID)
+		if err != nil {
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+
+		if user == nil || user.Role != s.role {
```
dpordomingo, line 35: this could be a good place to call GitHub API and ensure the user still satisfies app restrictions (irt)

src/api/index.js
```diff
@@ -103,10 +103,29 @@ function putAnswer(experimentId, assignmentId, answer) {
...
+
+function exportCreate() {
+  return apiCall(`/api/exports`, {
+    method: 'POST',
+  });
+}
+
+function exportDownload(filename) {
+  const token = TokenService.get();
+  const url = apiUrl(`/api/exports/${filename}/download?jwt_token=${token}`);
```
dpordomingo, line 118: but... is not the `jwt_token` something included in all requests?
```
authorization:Bearer header.payload.signature
```
why should it explicitly sent?

dpordomingo (commented):
LGTM... but there are some things that could be addressed at some point...

Issue #82, opened 2018-02-09
Fix message when no assignments available to an experiment


dpordomingo:
(context: https://github.com/src-d/code-annotation/pull/56#issuecomment-364480481)

It was strongly requested to configure the API to respond with an empty list when there are no assignments available to an experiment/user, but the frontend is not prepared for that change, so it is needed to fix it;

When the user has no assignments to fill, instead of rising a reasonable error, he obtains a "congrats for filling `NaN% experiments`, with no possibility to move to any other window.

![image](https://user-images.githubusercontent.com/2437584/36037627-a0710782-0dbd-11e8-8b7b-e0b272b7dc89.png)

PR #83, opened 2018-02-09
Fix NaN% on finish page

smacker:
Fixes https://github.com/src-d/code-annotation/issues/82

We don't have any other page to redirect a user. So, for now, user will see all zeros on finish page.


dpordomingo (approved):
LGTM,

bzz (approved):
:+1: for a test!

PR #84, opened 2018-02-09
Don't expose internal error

smacker:
Internal errors like errors from DB shouldn't leak to a user.

### DIFF
--- a/server/handler/render.go
+++ b/server/handler/render.go
@@ -47,7 +47,7 @@ func write(w http.ResponseWriter, r *http.Request, response *serializer.Response
 	} else {
 		statusCode = http.StatusInternalServerError
 		response.Status = statusCode
-		response.Errors = []serializer.HTTPError{serializer.NewHTTPError(statusCode, err.Error())}
+		response.Errors = []serializer.HTTPError{serializer.NewHTTPError(statusCode, http.StatusText(statusCode))}
 	}
 
 	if statusCode >= http.StatusBadRequest {


dpordomingo (commented):
Since **this is an internal tool**, that cannot be accessed by strange people, it may be a good idea to return the errors, and let the frontend to decide how to show them.
Again, returning errors in the API does not mean that they will be shown in the browser; they could be logged in the browser console instead.
The alternative of reading the pod logs from the google cloud engine is a pitty and it won't be easy to debug when we'll start having bugs to trace.

At least for the next days, we should be able to access the logs easily (especially while we're developing)

smacker:
no need to merge it right now, I agree. Let's keep this PR open until we sort out our infrastructure and then apply it.

dpordomingo:
What if we use an environment variable to toggle that behavior?
`KEEP_BACKEND_ERRORS=false`
Its default value can be `false` (hide logs), and in development and during the next days we can set it to `true` (keep errors). That feature would be only a couple of lines and would be very useful also in the future during development.
(and it could be also merged)

bzz:
Sorry for offtopic, but is that by any chance intentional that Drone is doing a CI and reports results in PRs? I belive it may be accidental and better be disabled.

smacker:
@dpordomingo for local development you have logs in console. For production usage, people usually use some errors aggregators. Like sentry, newrelic or just integrate error reporting to something like prometheus or whatever else.

dpordomingo (approved):
I don't like the idea of removing this useful info for an internal tool, but ok, let's merge it

bzz (approved):

PR #36, opened 2018-01-26
Single binary

smacker:
Fixes https://github.com/src-d/code-annotation/issues/34
Contains Gopkg.toml & Gopkg.yaml from https://github.com/src-d/code-annotation/pull/33

For easier understanding here is use cases:

For user:
  - CI react on a tag and run `make release` (no .travis config yet, we aren't ready for release)
  - It will produce single binary and docker image
  - binary will be uploaded to github for distribution

So, user needs:
  - download tarfile from releases
  - or use `docker run src-d/code-annotation`

For developers:

Global dependencies:
  - Go 1.9
  - Node 8
  - Yarn 1

Start contributing:
  - `go get -d -u github.com/src-d/code-annotation/...`
  - `$GOPATH/github.com/src-d/code-annotation`
  - `make serve`
  - no need for extra steps or commands before commit & push

For frontend developer who want to benefit from hot reloading:
  - `go get -d -u github.com/src-d/code-annotation/...`
  - `$GOPATH/github.com/src-d/code-annotation`
  - `make serve`
  - `make dev-frontend`

For crazy developers: (undocumented, unsupported but works)
  - `go get github.com/src-d/code-annotation/...` still works
  - but dependencies should be managed by the developer

PR #85, opened 2018-02-12
Prepare the project to be deployed

dpordomingo:
~~Depends on https://github.com/src-d/code-annotation/pull/36 Single Binnary~~
~~Depends on https://github.com/src-d/code-annotation/pull/91 Remove PR from Drone config~~
~~Depends on https://github.com/src-d/code-annotation/pull/90 Define a persistentVolume for internal database~~
~Depends on https://github.com/src-d/issues-infrastructure/issues/129 Push an example DB into the code-annotation volume~
~Depends on https://github.com/src-d/issues-infrastructure/issues/131 Push the final DB into the code-annotation volume~

Many different fixes and other additions; see commit messages :dagger:


helm-charts/code-annotation/templates/persistent-volume-claim.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  labels:
+    app: {{ template "fullname" . }}
+  name: {{ template "fullname" . }}
+spec:
+  accessModes:
+    - ReadWriteMany
```
rporres: not supported by GCE

helm-charts/code-annotation/values.yaml
```diff
@@ -16,6 +16,7 @@ service:
     externalPort: 8080
     internalPort: 8080
     name: code-annotation
+    internalDatabasePath: /var/code-annotation
```
rporres: This does not belong to the service

rporres (commented):
How does the application know that database path is in `/var/code-annotation`?

dpordomingo:
PR description updated.
@rporres your suggestions were addressed by the separated #90 Define a persistentVolume for internal database

bzz:
Updated PR description as src-d/issues-infrastructure#129  is done.

So, right now what's left before merging this seems to be:

- [x] fix the authentication problem (creating the OAuth app under srcd) 
- [ ] #36 to be merged
  - [ ] rebase over new master
  - [ ] sanity check
- [x] https://github.com/src-d/code-annotation/pull/85/commits/8d1b6a1978c915003ef2adcf9a7c28f9631995df to be removed

Please, feel free to update this comment in case something else is missing, etc.

bzz:
\cc @smacker for review

cli/server/server.go
```diff
@@ -16,7 +16,7 @@ type appConfig struct {
 	Host        string `envconfig:"HOST"`
 	Port        int    `envconfig:"PORT" default:"8080"`
 	UIDomain    string `envconfig:"UI_DOMAIN" default:"http://127.0.0.1:8080"`
-	DBConn      string `envconfig:"DB_CONNECTION" default:"sqlite://./internal.db"`
+	DBConn      string `envconfig:"DB_CONNECTION" default:"sqlite:///var/code-annotation/internal.db"`
```
smacker, line 19: by default server will fail because `/var/code-annotation` doesn't exist.
not very user-friendly.

Makefile
```diff
@@ -1,46 +1,98 @@
...
+# Tools
+YARN = yarn
+GODEP = dep
+GOLINT = golint
+GOVET = go vet
+BINDATA = go-bindata
+
+# ci variables
+TRAVIS_BUILD_DIR ?= $(shell pwd)
+CGO_ENABLED = 1
```
smacker: why?

Makefile
```diff
@@ -1,46 +1,98 @@
...
+	@git clone --quiet --depth 1 -b v1 $(CI_REPOSITORY) $(CI_PATH);
 -include $(MAKEFILE)
 
-# set enviroment variables from .env file
-include .env
-export $(shell sed 's/=.*//' .env)
+# Set enviroment variables from .env file
+ENV ?= .env
+$(ENV):
+	touch $(ENV)
```
smacker: I would prefer import `.env` only if it exists instead of creating new one. But up to you.

package.json
```diff
@@ -15,7 +17,7 @@
     "react-split-pane": "^0.1.74",
     "redux": "^3.7.2",
     "redux-devtools-extension": "^2.13.2",
-    "redux-little-router": "^14.2.3",
+    "redux-little-router": "^15.0.0",
```
smacker: could you describe what was the problem?

smacker (changes requested):
impossible to run locally with defaults

cli/server/server.go
dpordomingo: Since we're doing prod first that's the expected location we agreed.
We could maybe change the `.env.tpl` to be:
```shell
DB_CONNECTION=sqlite://internal.db
```

smacker: that's what I mean. It's absolutely okay to have `/var/code-annotation` path as default, but let's keep it simple for developers. Thanks!

smacker (approved):
good job! don't forget to squash commit later please.

cli/server/server.go
bzz: 👍  If anybody else wonders, feedback seems to be addressed and [default for dev env](https://github.com/src-d/code-annotation/pull/85/files#diff-4ea7d7d38652a9124d508ef1d16ac78dR5) was changed as requested.

bzz:
#36 is merged now, rebase is in order.

LGTM after that! 🎉

bzz (approved):

Makefile
dpordomingo: at some point, it was needed, but I also needed to change a big amount of things to make the project deployable.... and maybe another change made not necessary the CGO.
Removed

Makefile
dpordomingo: Done :dagger:

package.json
dpordomingo: While trying to make the project buildable it was needed at some point (it was somehow related to https://github.com/FormidableLabs/redux-little-router/pull/266), but it is not longer needed (idk why), so rollbacked.

dpordomingo:
rebased

PR #87, opened 2018-02-12
Improve assignments initialization

### LINKED ISSUE #75: There is no way to add assignments to a user that already have assignmnets
If it is run the `import` tool again, the file pairs are added to the DB, but users that already have assignments, won't be able to see those new file pairs

**Reason:**
because assignments are only created when they do not exist, not when they're new pairs to be labeled

carlosms:
Fixes #75, the problem when new file pairs are inserted into the DB, and the existing assignments do not pick them up.

### DIFF
--- a/server/repository/assignments.go
+++ b/server/repository/assignments.go
@@ -17,16 +17,26 @@ func NewAssignments(db *sql.DB) *Assignments {
 	return &Assignments{db: db}
 }
 
-// ErrNoAssignmentsInitialized is the error returned when the Assignments of a
-// User are requested for a given Experiment, but they have not been yet created
-var ErrNoAssignmentsInitialized = fmt.Errorf("No assignments initialized")
-
 const (
 	insertAssignmentsSQL = `INSERT INTO assignments (user_id, pair_id, experiment_id, answer, duration) VALUES ($1, $2, $3, $4, $5)`
-	selectIDFilePairsSQL = `SELECT id FROM file_pairs WHERE experiment_id=$1`
+	selectIDFilePairsSQL = `SELECT id FROM file_pairs WHERE experiment_id=$1 AND id NOT IN (SELECT pair_id FROM assignments WHERE experiment_id=$1 AND user_id=$2)`
 	selectAssignmentsSQL = `SELECT * FROM assignments WHERE user_id=$1 AND experiment_id=$2`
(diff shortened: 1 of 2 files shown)


bzz:
As there are no tests, would be nice to have a summary not only of the problem, but also of the expected behaviour.

I assume that if import is run again \w new file pairs, it is expected for all asgnments to pick up all new pairs?

carlosms:
The commits modify the behaviour in the following way.

When the assignments are requested for a certain experiment ID and user ID...

Before:
The server detects that there are 0 assignments, and the assignments are initialized from the existing file pairs. So from that moment, the assignments are considered initialized, and if new file pairs are inserted there will not be any new assignments for them.

Now:
The server checks if there are any file pairs without a related assignment. If so, new assignments are created, only for the missing file-pair IDs

smacker (approved):
good job!

bzz (approved):
Looks great to me.

server/repository/assignments.go
```diff
@@ -17,17 +17,27 @@ func NewAssignments(db *sql.DB) *Assignments {
...
-// User are requested for a given Experiment, but they have not been yet created
-var ErrNoAssignmentsInitialized = fmt.Errorf("No assignments initialized")
-
 const (
 	insertAssignmentsSQL = `INSERT INTO assignments (user_id, pair_id, experiment_id, answer, duration) VALUES ($1, $2, $3, $4, $5)`
-	selectIDFilePairsSQL = `SELECT id FROM file_pairs WHERE experiment_id=$1`
+	selectIDFilePairsSQL = `SELECT id FROM file_pairs WHERE experiment_id=$1 AND id NOT IN (SELECT pair_id FROM assignments WHERE experiment_id=$1 AND user_id=$2)`
 	selectAssignmentsSQL = `SELECT * FROM assignments WHERE user_id=$1 AND experiment_id=$2`
 	updateAssignmentsSQL = `UPDATE assignments SET answer='%v', duration=%v WHERE id=%v`
+	countPendingIDsSQL   = `SELECT count(id) FROM file_pairs WHERE experiment_id=$1 AND id NOT IN (SELECT pair_id FROM assignments WHERE experiment_id=$1 AND user_id=$2)`
```
bzz, line 25: By looking at these 2 SQL queries \w sub-queries, it makes me thing that if we have any more of those any time soon and need to change them - that would be a good argument for refactoring to use Postgres/Kallax.

PR #89, opened 2018-02-13
New endpoints for annotation results

### LINKED ISSUE #88: Endpoint for annotation results
for review page we need 2 more endpoints, something like:

`GET /exp/{id}/file-pairs` list of file pairs:
```
[
{id: 1, path: ... },
{id: 1, path: ... }
]
```

`GET /exp/{id}/file-pairs/{id}/annotations`

it should return json with user annotations, something like:
```
data: {
  yes: 1,
  no: 4,
  maybe: 2,
  skip: 0
}
```

carlosms:
Fixes #88.

There are two new routes:
- `experiments/{id}/file-pairs`
- `experiments/{id}/file-pairs/{id}/annotations`

Comments:
- Authorization (restrict to Requesters only) is left out of the scope of this PR, it will be addressed later.
- `serializer.simplifiedFilePairResponse` uses `leftPath` & `rightPath` to be consistent with `filePairResponse.left/rightBlobId`. But we may want to change it to a json with `left.Path` and `left.BlobId`


server/serializer/serializers.go
```diff
@@ -122,6 +138,22 @@ func NewFilePairResponse(fp *model.FilePair) *Response {
 		fp.ID, fp.Diff, fp.Score, fp.Left.BlobID, fp.Right.BlobID})
 }
 
+type simplifiedFilePairResponse struct {
```
smacker: do you think `listFilePairResponse` would be a better name?

server/repository/common.go
```diff
@@ -0,0 +1,6 @@
+package repository
+
+// scannable is used to call .Scan for both sql.Row and sql.Rows
+type scannable interface {
```
smacker, line 4: https://golang.org/pkg/database/sql/#Scanner ?

server/handler/assignments.go
```diff
@@ -96,3 +96,45 @@ func SaveAssignment(repo *repository.Assignments) RequestProcessFunc {
...
+		if err != nil {
+			return nil, err
+		}
+
+		pairID, err := urlParamInt(r, "pairId")
+		if err != nil {
+			return nil, err
+		}
+
+		assignments, err := repo.GetByExperimentPair(experimentID, pairID)
```
smacker, line 114: hm... it's not `assignments`, is it?

smacker (changes requested):
great work! just few small comments

server/repository/common.go
carlosms: I believe that is the interface used by Rows.Scan, to insert individual values into each argument.

server/handler/assignments.go
carlosms: Yes, those come from the assignments table

server/repository/common.go
smacker: Sorry, don't understand you. https://golang.org/pkg/database/sql/#Scanner is absolutely the same as this one.

server/handler/assignments.go
smacker: ah. got you. thanks!

server/repository/common.go
smacker: ah. it's actually different. 👍

smacker (approved):

bzz (approved):
LGTM

PR #90, opened 2018-02-13
Define a persistentVolume for internal database

dpordomingo:
This PR only prepares the volume where the sqlite should be `scp`ed (the app will be prepared in a separated PR)
As I was talking with @rporres we're not still supporting Postgresql in staging but sqlite.

After this PR is merged, it will be needed to push a sqlite bd into the volume as described by https://github.com/src-d/issues-infrastructure/issues/129

### DIFF
--- a/helm-charts/code-annotation/values.yaml
+++ b/helm-charts/code-annotation/values.yaml
@@ -10,6 +10,8 @@ image:
   repository: docker.io/srcd/code-annotation
   # tag must be received as a parameter
   pullPolicy: IfNotPresent
+deployment:
+    internalDatabasePath: /var/code-annotation
 service:
   type: NodePort
   codeAnnotation:
--- a/helm-charts/code-annotation/templates/deployment.yaml
+++ b/helm-charts/code-annotation/templates/deployment.yaml
@@ -17,6 +17,10 @@ spec:
         app: {{ template "name" . }}
         release: {{ .Release.Name }}
     spec:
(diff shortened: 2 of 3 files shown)


helm-charts/code-annotation/values.yaml
```diff
@@ -10,6 +10,8 @@ image:
   repository: docker.io/srcd/code-annotation
   # tag must be received as a parameter
   pullPolicy: IfNotPresent
+deployment:
+    internalDatabasePath: /var/code-annotation
```
rporres, line 14: If there's no env variable that will inform that this may change beyond the default, I would not give the possibility of overriding it and I would hardcode it in the deployment yaml template

rporres (changes requested):

helm-charts/code-annotation/values.yaml
dpordomingo: I think this https://github.com/src-d/code-annotation/pull/90/commits/bfa612c1f6fa632f0ca3ef2c39f158d4c0d8bf12 is what you're looking for... isn't it?

rporres (approved):

dpordomingo:
You're right, thanks for spotting it @rporres 
fixed? https://github.com/src-d/code-annotation/pull/90#discussion_r167927792

bzz (approved):
LGTM

smacker (approved):

PR #94, opened 2018-02-13
Connect review page to backend

smacker:
Partly solves https://github.com/src-d/code-annotation/issues/47

Based on https://github.com/src-d/code-annotation/issues/88

PR #95, opened 2018-02-13
Update browser title for each page

### LINKED ISSUE #92: Correct page titles
Currently, all pages have a title "React App".
We need to set correct titles depends on a page.
Something like:
```
index: Source{d} Code Annotation Tool
experiment: <name> | Source{d} Code Annotation Tool
review: Review for experiment <name> | Source{d} Code Annotation Tool
```

carlosms:
Fixes #92 

Review.js does not work for now, but I hope it works after #94 is merged.

### DIFF
--- a/src/pages/Experiment.js
+++ b/src/pages/Experiment.js
@@ -32,7 +32,7 @@ class Experiment extends Component {
     return (
       <div className="ex-page">
         <Helmet>
-          <title>{`${this.props.name} | source{d} Code Annotation Tool`}</title>
+          <title>{this.props.name}</title>
         </Helmet>
         <PageHeader {...user} />
         {this.renderMain()}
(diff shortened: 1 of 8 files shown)


smacker:
hm... Do you think it would be better to use https://github.com/nfl/react-helmet ?
it would allow to use templates for titles and instead of juggling with componentDidMount/componentWillReceiveProps you can just put it in render function.

also finish page is missed.

carlosms:
Thanks @smacker, I've pushed new code using react-helmet.

src/pages/Experiment.js
```diff
@@ -30,6 +31,9 @@ class Experiment extends Component {
 
     return (
       <div className="ex-page">
+        <Helmet>
+          <title>{`${this.props.name} | source{d} Code Annotation Tool`}</title>
```
smacker: please use `titleTemplate` to avoid copy-pasting suffix.

smacker (changes requested):

dpordomingo (approved):

bzz (approved):
LGTM

smacker (approved):

bzz:
@carlosms is there anything left before merging this guy?

PR #100, opened 2018-02-14
Restrict access to org:src-d members

dpordomingo:
Only members of `src-d` will be accepted as valid users.
(the org can be configured during the deploy)

### DIFF
--- a/helm-charts/code-annotation/templates/deployment.yaml
+++ b/helm-charts/code-annotation/templates/deployment.yaml
@@ -26,6 +26,10 @@ spec:
           image: "{{ .Values.image.repository }}:{{ required "Image tag is required" .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           env:
+            - name: OAUTH_RESTRICT_ACCESS
+              value: "{{ .Values.authorization.restrictAccessGroup }}"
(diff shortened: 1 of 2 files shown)


rporres (approved):

bzz (approved):
LGTM.

:+1: Empty value for `restrictRequesterGroup ` looks like a [good default](https://github.com/src-d/code-annotation/commit/8b3174e6a55e8e0e08221d94a72dc60f6a6c3c6d#diff-96702da8eec3f188048d4e6658fd4494R111)

rporres (commented):
Just a detail. To be on the safe side, always quote whatever value for an env variable you introduce

carlosms:
I think that if we do it this way, all users will be created with the `worker` role. And we don't have a way (yet) to change an existing user role.

Why don't we create two teams now, and create accounts properly as `requester`?

dpordomingo:
@rporres I quoted 4a034c6 the values as you suggested 

@carlosms is there any (real-big) problem if all of us can access the review panels? I'd not block the PR because of that... and if at some point we should hide the review panels, we can change the auth model so it would be done in real time, instead of on user creation (otherwise, we won't be able to un-authorise a user)

dpordomingo:
@rporres done for tpls 24f75c0 too :dagger:

rporres (approved):

Issue #97, opened 2018-02-14
Log overflow


dpordomingo:
Could we avoid logs from internal things?

![image](https://user-images.githubusercontent.com/2437584/36191433-6e944582-115d-11e8-8239-7161bf5fc2af.png)

bzz:
Indeed, we should setup default log level to WARN at least for staging deployment.

dpordomingo:
Mmmmm... @bzz I didn't understand how your message https://github.com/src-d/code-annotation/issues/97#issuecomment-365550835 is related with the issue desc :upside_down_face:

dpordomingo:
https://github.com/src-d/code-annotation/pull/102 is not related with this issue

smacker:
what? PR fixes it:
> we should setup default log level to WARN at least for staging deployment

no more "overflow" of INFO logs

dpordomingo:
Maybe the issue desc is not clear enough, but it is said 
>_Could we avoid logs from internal things?_

And a screenshot with tons of automatic access by IP

smacker:
hahaha. I got it as we should remove `INFO` messages. And looks like @bzz also.

bzz:
I think further discussion IRL leaked to https://github.com/src-d/code-annotation/pull/36#issuecomment-365628856

dpordomingo:
logging specs were described by https://github.com/src-d/code-annotation/issues/123
We'll track that development in the new issue

PR #102, opened 2018-02-14
Improve logging

smacker:
Fixes https://github.com/src-d/code-annotation/issues/97

* Set log level depends on env and pass logger around correctly
* New vendor dependency


bzz (approved):
LGTM.

Rebase seems to be in order though.

server/service/logger.go
```diff
@@ -5,12 +5,16 @@ import (
...
 // NewLogger returns a logrus Logger
-func NewLogger() logrus.FieldLogger {
+func NewLogger(env string) *logrus.Logger {
 	logger := logrus.New()
 	logger.Formatter = &logrus.TextFormatter{
 		FullTimestamp:   true,
 		TimestampFormat: "2006-01-02 15:04:05",
 	}
 
+	if env == "production" || env == "staging" {
```
dpordomingo: `production` behavior should be the default one.
It's better to override things when we're in development.

dpordomingo (changes requested):
mostly lgtm but one thing about the defaults...

server/service/logger.go
smacker: I mimic `yarn/npm` behavior. But production by default is also ok for me. @bzz what do you think?

dpordomingo: We're changing things to make the app "prod first"... so I'm not sure if it's a good idea to develop new things in the other way...

bzz: 👍  for production-first approach

smacker:
changed to prod first approach

dpordomingo (approved):
LGTM....

I hope you only changed this https://github.com/src-d/code-annotation/pull/102/files#diff-3e5785ed7dfffb946332460b8a47b99dR15
Otherwise... idk

(commits in PRs being reviewed should not be amended; otherwise, the reviewer can not review (only) the changes)

smacker:
Default has been changed from `dev` to `production` and env variable is removed from helm chart.
If you skip vendoring -  there are just 16 lines of code have been changed in total (yes, I counted).

bzz:
👍  Shall we merge this guy then?

PR #103, opened 2018-02-14
Fix bug in OAuth

### LINKED ISSUE #101: Wrong logic for role assignment with oauth_restrict_* strings
In OAuth.GetUser: Right now if restrictRequesterAccess is empty, restrictAccess is not checked.

carlosms:
Part of the #101

### DIFF
--- a/server/service/oauth.go
+++ b/server/service/oauth.go
@@ -107,24 +107,22 @@ func (o *OAuth) GetUser(ctx context.Context, code string) (*githubUser, error) {
 		return nil, fmt.Errorf("can't parse github user response: %s", err)
 	}
 
-	role := model.Requester
+	user.Role = model.Requester
 	if o.restrictRequesterAccess != "" {
 		err := o.checkAccess(client, o.restrictRequesterAccess, user.Login)
 		if err != nil && err != ErrNoAccess {
 			return nil, err
 		}
 		if err == ErrNoAccess {
-			role = model.Worker
+			user.Role = model.Worker
 		}
 	}
 
-	if o.restrictAccess != "" && role == model.Worker {
+	if o.restrictAccess != "" && (user.Role == model.Worker || o.restrictRequesterAccess == "") {
 		if err := o.checkAccess(client, o.restrictAccess, user.Login); err != nil {
 			return nil, err
 		}
 	}
-
-	user.Role = role
 
 	return &user, nil
 }


carlosms:
Now the logic should work like this:

| RESTRICT_ACCESS | RESTRICT_REQUESTER_ACCESS | Anyone | User in acc_group | User in req_group |
|--|--|--|--|--|
| - | - | Requester | N/A | N/A |
| - | req_group | Worker | N/A | Requester |
| acc_group | - | Denied | Requester | N/A |
| acc_group | req_group | Denied | Worker | Requester |

smacker:
Looks good, but I would love to have tests for it.

server/service/oauth.go
```diff
@@ -107,25 +107,23 @@ func (o *OAuth) GetUser(ctx context.Context, code string) (*githubUser, error) {
...
-	role := model.Requester
+	user.Role = model.Requester
 	if o.restrictRequesterAccess != "" {
 		err := o.checkAccess(client, o.restrictRequesterAccess, user.Login)
 		if err != nil && err != ErrNoAccess {
 			return nil, err
 		}
 		if err == ErrNoAccess {
-			role = model.Worker
+			user.Role = model.Worker
```
bzz, line 117: A bit confused, please, help me understand what is going to happen in case
 - if `restrictRequesterAccess=something` 
 - but this user is from a different org (`err == ErrNoAccess` ?)
 - and what is going to happen in case [next if's condition below](https://github.com/src-d/code-annotation/pull/103/files#diff-96702da8eec3f188048d4e6658fd4494R121) is false (i.e `restrictAccess==""` )?

bzz:
I belive test are a must in this case.

@dpordomingo @carlosms what do you think if merge this ASAP and ask @carlosms to refactor/add tests for each case from [an awesome table above](https://github.com/src-d/code-annotation/pull/103#issuecomment-365599941) in a separate PR?

server/service/oauth.go
carlosms: The user is not part of the requester's group, so they are assigned the `Worker` role. The next `if` condition is false, so the user is considered authorized.
This corresponds with the table row from the comment above:

| RESTRICT_ACCESS | RESTRICT_REQUESTER_ACCESS | Anyone | User in acc_group | User in req_group |
|--|--|--|--|--|
| - | req_group | Worker | N/A | Requester |

smacker (dismissed):
need tests

dpordomingo:
I find the original code complex and not obvious, so I'd propose this alternative... 
feel free to proceed as you want
```go

defaultRole := model.Requester

if o.restrictRequesterAccess != "" {
	if err := o.checkAccess(client, o.restrictRequesterAccess, user.Login); err == nil {
		user.Role = model.Requester
		return &user, nil
	} else if err != ErrNoAccess {
		return nil, err
	}

	defaultRole = model.Worker
}

if o.restrictAccess != "" {
	if err := o.checkAccess(client, o.restrictAccess, user.Login); err != nil {
		return nil, err
	}

	user.Role = defaultRole
	return &user, nil
}

user.Role = defaultRole
return &user, nil
```

dpordomingo (approved):
I'd suggest https://github.com/src-d/code-annotation/pull/103#issuecomment-365614878 but LGTM

bzz:
Let's merge, keep #101 open and refactor to simplify logic+add tests \w mock in subsequent PR to cover all conditions from https://github.com/src-d/code-annotation/pull/103#issuecomment-365599941

bzz (approved):
LGTM, given the plan discussed above

PR #108, opened 2018-02-14
Test for #101

carlosms:
Tests for #101.

It also contains the code modification suggested by @dpordomingo  here: https://github.com/src-d/code-annotation/pull/103#issuecomment-365614878


server/service/oauth_test.go
```diff
@@ -0,0 +1,115 @@
...
+TestGetUser tests role assignment based on the group membership. The tested
+logic is the following table:
+
++-------------------+-----------------------------+-------------+---------------+------------------+
+|                   |                             |             |  User in      |  User in         |
+|  RESTRICT_ACCESS  |  RESTRICT_REQUESTER_ACCESS  |  Outsider   |  accessGroup  |  requesterGroup  |
++-------------------+-----------------------------+-------------+---------------+------------------+
+|  -                |  -                          |  Requester  |  N/A          |  N/A             |
+|  -                |  requesterGroup             |  Worker     |  N/A          |  Requester       |
+|  accessGroup      |  -                          |  Denied     |  Requester    |  N/A             |
```
dpordomingo, line 81: why not `Denied` in 3rd col?

server/service/oauth_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+/*
+TestGetUser tests role assignment based on the group membership. The tested
+logic is the following table:
+
++-------------------+-----------------------------+-------------+---------------+------------------+
+|                   |                             |             |  User in      |  User in         |
+|  RESTRICT_ACCESS  |  RESTRICT_REQUESTER_ACCESS  |  Outsider   |  accessGroup  |  requesterGroup  |
++-------------------+-----------------------------+-------------+---------------+------------------+
+|  -                |  -                          |  Requester  |  N/A          |  N/A             |
```
dpordomingo, line 79: why not `Requester` in 2nd and 3rd cols?

server/service/oauth.go
```diff
@@ -107,24 +107,45 @@ func (o *OAuth) GetUser(ctx context.Context, code string) (*githubUser, error) {
 		return nil, fmt.Errorf("can't parse github user response: %s", err)
 	}
 
-	user.Role = model.Requester
+	if err = o.setRole(o, client, &user); err != nil {
```
smacker: `o.setRole(o)` looks very weird for me.
Do you think you could refactor it like this:

```go
type OAuth struct {
    // ...
	restrictionChecker restrictionChecker
}

func NewOAuth(clientID, clientSecret, restrictAccess, restrictRequesterAccess string) *OAuth {
	// ...
	return &OAuth{
        // ...
		restrictionChecker:      &githubPermissions{}
	}
}

func (o *OAuth) setRole(client *http.Client, user *githubUser) error {
	//...
	o.restrictionChecker(client, o.restrictAccess, user.Login)
	// ...
}

type githubPermissions struct {}

func (c *githubPermissions) checkAccess(client *http.Client, restriction, login string) error {
	// ...
}
```

or maybe you see how to do it even better?

smacker (approved):
LGTM. But could you please consider my suggestion?

bzz:
@smacker suggestion makes a lot of sense, what do you think @carlosms ?

carlosms:
@smacker thanks for the suggestion, now it looks cleaner.
@bzz please take another look.

server/service/oauth_test.go
```diff
@@ -0,0 +1,124 @@
...
+|  -                |  requesterGroup             |  Worker     |  N/A          |  Requester       |
```
dpordomingo, line 80: why not `Worker` in 2nd col?

dpordomingo (approved):
LGTM but one thing in the table

server/service/oauth.go
```diff
@@ -107,30 +114,49 @@ func (o *OAuth) GetUser(ctx context.Context, code string) (*githubUser, error) {
...
 
-	if o.restrictAccess != "" && (user.Role == model.Worker || o.restrictRequesterAccess == "") {
-		if err := o.checkAccess(client, o.restrictAccess, user.Login); err != nil {
-			return nil, err
+	if o.restrictAccess != "" {
+		if err := o.restrictionChecker.checkAccess(client, o.restrictAccess, user.Login); err != nil {
+			return err
 		}
+
+		user.Role = defaultRole
```
dpordomingo: I think this and the following one are redundant; not necessarily because they're repeated outside https://github.com/src-d/code-annotation/pull/108/files#diff-96702da8eec3f188048d4e6658fd4494R149

smacker (approved):
🎉

server/service/oauth_test.go
carlosms: Because that column refers to the users members of the group set in in RESTRICT_REQUESTER_ACCESS. If there isn't one set, you can't reason about the users belonging to that group. Hence, N/A.

Does it make sense?

server/service/oauth.go
carlosms: Yeah, I just copied your code as suggested here https://github.com/src-d/code-annotation/pull/103#issuecomment-365614878

I can remove these lines if you tell me to, I though you left them intentionally to make it more legible.

server/service/oauth_test.go
dpordomingo: I understood from your explanation irl that the table was the key to understand the testcases below; if so, I think it would be useful to have a complete table.
For example, from the second row:
> if the **requester rule is defined,** and the user trying to access is in the **group of people that can access** the app, the assigned role can be **whatever**

And it does not match with the real test case, that expects the **worker** role for that user.

server/service/oauth.go
dpordomingo: mine was only compilable pseudocode :wink: but whatever decision you take will be fine for me :dancer:

server/service/oauth_test.go
carlosms: I have just pushed e8673cba1b4b16a9c9aaf11bdbb9b3cee7cede0e to add the N/A option as a test result.

By the way, just to clarify any confusion, the second row does not mean:
> if the **requester rule is defined,** and the user trying to access is in the **group of people that can access** the app, the assigned role can be **whatever**

```
+-------------------+-----------------------------+-------------+---------------+------------------+
|                   |                             |             |  User in      |  User in         |
|  RESTRICT_ACCESS  |  RESTRICT_REQUESTER_ACCESS  |  Outsider   |  accessGroup  |  requesterGroup  |
+-------------------+-----------------------------+-------------+---------------+------------------+
|  -                |  requesterGroup             |  Worker     |  N/A          |  Requester       |
```

In this case there isn't a _group of people that can access the app_. There are only users on the `requesterGroup`, and people that are not in it. Members of the `accessGroup`do not hold any special meaning in this test case, because they are not part of the restriction set in `RESTRICT_ACCESS`, and the test is _not applicable_.
But it does not mean _whatever_, of course. As any other user not in the `requesterGroup`, these users fall into the outsider column, that is, `worker` role.

bzz (approved):
LGTM!

But there still seems to be some questions from @dpordomingo left

dpordomingo (approved):

server/service/oauth_test.go
dpordomingo: thanks!

server/service/oauth_test.go
dpordomingo: solved by e8673cb

server/service/oauth_test.go
dpordomingo: solved by e8673cb

Issue #114, opened 2018-02-15
Add version number


bzz:
Add version number based on Git information to the build and expose it through http endpoint.

PR #115, opened 2018-02-15
Add version endpoint

smacker:
Fixes: https://github.com/src-d/code-annotation/issues/114

### DIFF
--- a/cli/server/server.go
+++ b/cli/server/server.go
@@ -11,6 +11,8 @@ import (
 	"github.com/kelseyhightower/envconfig"
 )
 
+var version = "dev"
+
 type appConfig struct {
 	Env         string `envconfig:"ENV" default:"production"`
 	Host        string `envconfig:"HOST"`
@@ -56,7 +58,7 @@ func main() {
 	jwt := service.NewJWT(jwtConfig.SigningKey)
 
 	// start the router
-	router := server.Router(logger, jwt, oauth, conf.UIDomain, &db, "build", conf.ExportsPath)
+	router := server.Router(logger, jwt, oauth, conf.UIDomain, &db, "build", conf.ExportsPath, version)
 	logger.Info("running...")
 	err = http.ListenAndServe(fmt.Sprintf("%s:%d", conf.Host, conf.Port), router)
 	logger.Fatal(err)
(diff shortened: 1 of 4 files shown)


cli/server/server.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/kelseyhightower/envconfig"
 )
 
+var version = "dev"
```
bzz: What do you think, if instead of having it hard-coded we replace it at the build time with something like `git describe` using `-ldflags`

smacker: https://github.com/src-d/ci/blob/v1/Makefile.main#L56

this one is default value when you do `go run` for example.

dpordomingo (approved):
I don't know what'd be the use case of this endpoint, but it's formally ok, so LGTM

cli/server/server.go
carlosms: Suggestion/request: to make it easier for new team members, it would be nice to have your explanation as a comment above the variable, like:
`version will be replaced automatically by the CI build. See <link>`

smacker: oh come on. it's how all src-d go projects work:

https://github.com/src-d/enry/blob/master/cli/enry/main.go#L20
https://github.com/src-d/borges/blob/master/cli/borges/main.go#L19
https://github.com/src-d/go-git/blob/master/cli/go-git/version.go#L5
https://github.com/bblfsh/bblfshd/blob/master/cli/bblfshd/main.go#L20
https://github.com/bblfsh/dashboard/blob/master/server/cmd/bblfsh-dashboard/main.go#L20

bzz: Huge 👍 on what @carlosms says. Not necessarily the comment in code, but at least on a PR.

@smacker thank you for the links!

Caring about new contributors, including src-d team members, is very aligned with https://github.com/src-d/code-annotation/blob/master/CODE_OF_CONDUCT.md#our-standards

> Showing empathy towards other community members

bzz (approved):

cli/server/server.go
smacker: put comments about CI in the __code__? really?
what next? describe our kubernetes configuration in the code?
comment that algorithm is going to work fast because we run it on science3 with GPU?

whatever, I added a comment

PR #124, opened 2018-02-15
Add role restrictions to routes

carlosms:
Part of #47. This PR adds authorization restrictions based on the user role.

In the backend the server replies with an unathorized response.
In the frontend a restricted route redirects to a new page, `unauthorized`, that shows a simple information message.

127 0 0 1_3000_export

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -8,6 +8,7 @@ import Experiment from './pages/Experiment';
 import Final from './pages/Final';
 import Review from './pages/Review';
 import Export from './pages/Export';
+import Forbidden from './pages/Forbidden';
 
 class App extends Component {
   render() {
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -8,6 +8,7 @@ import Experiment from './pages/Experiment';
 import Final from './pages/Final';
 import Review from './pages/Review';
 import Export from './pages/Export';
+import Unauthorized from './pages/Unauthorized';
```
smacker: it might be nitpick but `Unauthorized` usually means a little bit different thing (get unauthorized = logged out).
Could you please rename it to `Forbidden` please? Also, it would be more consistent with backend where we use `401 Unauthorized` to logout user and `403 Forbidden` to disallow to see protected content.

carlosms: Yes, it makes sense, I've changed it in 5b2d777

smacker (approved):
🎉

bzz (approved):
LGTM

PR #128, opened 2018-02-16
Update user data after each log in

### LINKED ISSUE #116: Role update: should happen on login
Right now role is assigned to a user only once. 

But to be able to change roles for users, it would be great to check \w GH if the role has been changed and store it in DB on every login.

carlosms:
Fixes #116


bzz (approved):
LGTM

dpordomingo (approved):

PR #141, opened 2018-02-20
Test build in travis

### LINKED ISSUE #125: Add build check on Travis
Travis is now only running lint and test.
It can happen (it happened) that the build will fail in drone after the PR is merged -> what could be a big problem (edited)
We should add to the Travis rules the building of the app, so we'll notice the app is not buildable in advance (before merging)

dpordomingo:
fix #125 

once this is merged, Travis will check if project is buildable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,7 @@ jobs:
       script:
         - make lint-frontend
         - make test-frontend
-        - make dependencies-backend
-        # uncommend next line when BE code will start passing go lint and go vet
-        # - make lint-backend
+        - make lint-backend
         - make test
         - make validate-commit
     - script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,16 @@ before_install:
   - nvm use 8
   - npm install -g yarn
 
-script:
-  - make lint-frontend
-  - make test-frontend
-  - make dependencies-backend
-  # uncommend next line when BE code will start passing go lint and go vet
-  # - make lint-backend
-  - make test
-  - make validate-commit
+jobs:
+  include:
+    - stage: Tests, linters and build
+      script:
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -9,10 +9,13 @@ before_install:
...
+jobs:
+  include:
+    - stage: Tests, linters and build
+      script:
+        - make lint-frontend
+        - make test-frontend
+        - make lint-backend
+        - make test
+    - script:
+        - make build-app
```
smacker: no need to create tars for testing.

dpordomingo: The issue requires to test the whole building process

smacker: creating tars isn't building.

dpordomingo: Just called `make packages` and whatever it does, is ok for the purpose of the issue.
Is there any problem if the tars are generated?

smacker: nothing really wrong. just my perfectionism 😅 
Task has the name `test build`. Comment in travis says `build`. And we have a target `make build` exactly for this purpose - to build.

dpordomingo: Where is the name `test build`?

aside:
`build` rule is called from `packages` rule, that is called from `build-app` that is what drone really calls, that is what we're trying to reproduce.

smacker: I meant the name of the job:`stage: Tests, linters and build`.

As you may notice I didn't request changes, it's okay to merge it as it is.

bzz:
@dpordomingo rebase seems to be in order.

Also I'm confused - review requested, but #125 is not in "Code Review" state according to the Kanban.

bzz (approved):
LGTM, sans https://github.com/src-d/code-annotation/pull/141#issuecomment-367266375

PR #142, opened 2018-02-20
Handle auth in front end

### LINKED ISSUE #99: Handle internal server errors in frontend.
There are some URLs that can return a `500` error without being caught by the frontend, so the site will show a meaningless plain error message like the following one received from [/oauth-callback](https://code-annotation-staging.srcd.run/oauth-callback) when GitHub auth fails 

The frontend should be able to catch those errors.

image

dpordomingo:
Fix #99

major changes:
- new route on the front side: `/auth`:
  - it will receive the `state` and `code` from  GitHub OAuth callback,
  - it will call through AJAX the (rename) backend `/api/auth` endpoint, and handle any possible error,
- `UI_DOMAIN` is no longer needed in Helm charts

benefits:
- the backend `/api/auth` endpoint will be standard in terms of our `handler.RequestProcessFunc`,
- the backend will communicate with the front end as the rest of the app:<br />FE &rarr; BE &rarr; response + status

possible improvements:
- I realized that `/auth` route is not really-really needed because the auth thing could be handled by `/` itself if `code` is received.

### DIFF
--- a/server/service/oauth.go
+++ b/server/service/oauth.go
@@ -68,16 +68,22 @@ type GithubUser struct {
 }
 
 // MakeAuthURL returns string for redirect to provider
-func (o *OAuth) MakeAuthURL(w http.ResponseWriter, r *http.Request) string {
+func (o *OAuth) MakeAuthURL(w http.ResponseWriter, r *http.Request) (string, error) {
 	b := make([]byte, 16)
 	rand.Read(b)
 	state := base64.URLEncoding.EncodeToString(b)
 
-	session, _ := o.store.Get(r, "sess")
-	session.Values["state"] = state
-	session.Save(r, w)
+	session, err := o.store.Get(r, "sess")
+	if err != nil {
+		return "", fmt.Errorf("could not save state under the current session: %s", err)
+	}
 
-	return o.config.AuthCodeURL(state)
+	session.Values["state"] = state
+	if err := session.Save(r, w); err != nil {
+		return "", fmt.Errorf("could not save state under the current session: %s", err)
+	}
+
+	return o.config.AuthCodeURL(state), nil
 }
 
 // ValidateState protects the user from CSRF attacks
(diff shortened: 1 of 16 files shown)


smacker (changes requested):
1. You can't read POST payload in javascript

So your idea FE->BE->reponse doesn't work.

2. You don't handle errors from provider:
https://tools.ietf.org/html/rfc6749#section-4.1.2.1

And we have to provide those errors to the user because those are problems between user&provider, not our app.

dpordomingo:
- There is no post payload to read,
- errors during auth phase are returned to the FE as usual.

dpordomingo:
It must be rebased over master and refactored considering https://github.com/src-d/code-annotation/pull/144

dpordomingo:
PR rebased over current `master`.

This PR should not be merged till it can be coordinated with https://github.com/src-d/code-annotation/issues/201

PR now includes changes proposed by @smacker at https://github.com/src-d/code-annotation/pull/150:
- 35d2f5b refactoring 
- 0e95b7a Remove redundant UI_DOMAIN 
- a3f08a7 Fix new test from master

It was also included some minors:
- 52b7f45 Log OAuth errors errors

server/service/oauth.go
```diff
@@ -68,16 +68,18 @@ type GithubUser struct {
 }
 
 // MakeAuthURL returns string for redirect to provider
-func (o *OAuth) MakeAuthURL(w http.ResponseWriter, r *http.Request) string {
+func (o *OAuth) MakeAuthURL(w http.ResponseWriter, r *http.Request) (string, error) {
 	b := make([]byte, 16)
 	rand.Read(b)
 	state := base64.URLEncoding.EncodeToString(b)
 
 	session, _ := o.store.Get(r, "sess")
```
smacker: if you return `(string, error)`, it makes sense to handle error here too.

smacker (changes requested):
Oh. You cherry-picked my commits. Okay.

One small comment about error handling.

carlosms (approved):

server/service/oauth.go
dpordomingo: yesssss, you're totally right; thanks for spotting it!

dpordomingo:
Ready for a second pass @bzz @smacker 
- 3f6e7f5 error handling
- 737ce23 use `/oauth-callback` as requested by https://github.com/src-d/code-annotation/issues/201

smacker (approved):

Issue #140, opened 2018-02-19
Add simple analytics


bzz:
To be able to grow this project UX based on it's actual usage a designer would need to have access to some data.

This issue includes:
 - enable Google Analytics conditionally (dev env should not have one, staging deployment should have one, production would have different one) 
 - as a start, only emit [PageView events](https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications) for every URL change in our FE

ricardobaeta:
It would be great to know which UI element triggered the URL change.

dpordomingo:
Yes @ricardobaeta it makes sense and is doable, but I think it should be tracked it in a separated issue, so if @bzz agree, I'd suggest you open a new one.

bzz:
Analytics was added, subsequent work will be done in separate PRs.

PR #143, opened 2018-02-20
Pass env variables to frontend

smacker:
These changes allow to pass environmental variables to frontend.

- It makes images more flexible (we don't set server URL inside image anymore)
- We can change frontend behavior depends on environment now
- Server can pass initial state to frontend if needed


carlosms (approved):
LGTM

bzz (approved):
LGTM

This also should enable flexible GA token passage to FE in #140

Issue #146, opened 2018-02-20
Single scrollbar


vmarkovtsev:
The current diff view has two separate scrollbars - per each file. This is undesirable for the whole ML team. Actually, we thought that Firefox displays it correctly and Chrome has a bug. We request to have a single scrollbar.

vmarkovtsev:
Thanks a lot!

PR #147, opened 2018-02-20
Single scrollbar

smacker:
Fixes: https://github.com/src-d/code-annotation/issues/146

before
image

After
image

image

### DIFF
--- a/src/components/Experiment/Diff.js
+++ b/src/components/Experiment/Diff.js
@@ -4,6 +4,19 @@ import 'diff2html/dist/diff2html.css';
 import './Diff.less';
 
 class Diff extends PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.leftDiff = null;
+    this.rightDiff = null;
+
+    this.leftListen = true;
+    this.rightListen = true;
+
+    this.leftScrollListener = this.listerner('left').bind(this);
+    this.rightScrollListener = this.listerner('right').bind(this);
+  }
+
   render() {
     const { diffString, leftLoc, rightLoc, className } = this.props;
     const diffHTML = Diff2Html.getPrettyHtml(diffString, {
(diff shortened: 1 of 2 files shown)


bzz:
@smacker so both, FF and Chrome are tested and have the small behaviour of a single.scroll-bar for a file pair now?

smacker:
Yes. Both have same behavior of single scroll bar. That is what issue & PR are about, no?

dpordomingo (commented):
As it can be seen in the screenshot, horizontal scrollbars are cropped in Chrome.
But what might be worst is that to do horizontal scroll, the user needs to scroll to the bottom of the page because that's the new position of the horizontal scrollbars.
Is that behavior expected for the users?

dpordomingo:
Please, @smacker add screenshots to the PR descriptions when the changes have an impact on the UI.
I already did it to this PR because I needed them to provide my feedback, but it would be useful if it is done by the one who creates the PR :wink:

smacker:
@dpordomingo hm. On mac os scrollbars are ok. But I'll try to check and fix (will need test from somebody on linux though later anyway). Thanks!

smacker:
@dpordomingo there are no visual changes on mac os compare to master.

dpordomingo:
The thing that concerns me more is the necessity of scroll to the bottom to do h-scrolling; I think it should be approved by @ricardobaeta and/or our users to ensure it fits its expectations

bzz:
Thank you for rising this, @dpordomingo !

> the necessity of scroll to the bottom to do h-scrolling

I'm sure @vmarkovtsev as the author of #146 will be happy to provide input on this here ASAP

bzz (approved):
LGTM

as soon as we know if https://github.com/src-d/code-annotation/pull/147#issuecomment-367268179 is acceptable for users or find a better solution

smacker:
> The thing that concerns me more is the necessity of scroll to the bottom to do h-scrolling

makes sense. I don't use a mouse, so I didn't even think about it. Please give me a bit of time. I'll try to fix cropped scrollbar on non mac os and this issue.

dpordomingo (approved):
LGTM, 
@smacker you can merge

carlosms (approved):

smacker:
For the history: on Mac OS in FF it works but the feeling of the scroll is broken. On Linux it works well in both browsers.
We decided it's good enough for now. We will ask Mac OS users to switch to Chrome.

PR #150, opened 2018-02-21
Authorization refactoring

smacker:
Fixes https://github.com/src-d/code-annotation/issues/99

Thanks @dpordomingo for good example which I could test and implement it a little bit "righter" way:

- Keep frontend logic for auth out of components
- Make UI more similar to how it works right now (all errors go to popup) and user is able to retry login again.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -91,7 +91,7 @@ If you want to benefit from frontend hot reloading feature:
 Run server. Execute:
 
 ```bash
-$ UI_DOMAIN=http://127.0.0.1:3000 make gorun
+$ make gorun
 ```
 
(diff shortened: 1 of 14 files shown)


carlosms (approved):
Looks good to me.

My only concern is that we need to coordinate with the infra. team to change the current GitHub application callback, and this might be easy to forget.

Maybe we should open a new issue for the 0.0.3 release with a comment to remind us of this (ping @dpordomingo)

smacker:
Good point about callback. Only it should be coordinated with admin of the application. (I'm not sure it's infra)

dpordomingo:
superseded by original PR https://github.com/src-d/code-annotation/pull/142

Issue #149, opened 2018-02-21
Add multiple experiments support


bzz:
There is going to be 1 more dataset to label coming soon, of 500 more file pairs.

To prepare for it, we need to:

 - [x] add frontend experiment list page (without real data)
 - [x] make sure backend (api) works correctly with more than 1 experiment (we *might* have something hardcoded)
 - [x] add support for multiple experiments on frontend side (many places assume only one or have hardcoded value now)
 - [x] update importing tools, so the data from [sqlite format](https://github.com/src-d/code-annotation/blob/master/cli/import/examples/example.sql) works for multiple experiments as expected (currently, there is hardcoded experiment 1)

smacker:
done!

PR #154, opened 2018-02-21
Add new route /experiments

carlosms:

The new route lists all the existing experiments for logged-in users.


smacker (approved):
Thanks!

bzz (approved):
LGTM

Sans the fact that from application standpoint it still might make sense to:
 - return a partial results from DB, in case of failure
 - have a `LIMIT N` with some big `N` in all select queries

both of which do not block the merge and might need to be consistently applied in a different PRs.

PR #152, opened 2018-02-21
Implement page with a list of experiments

smacker:
Implements the first point in https://github.com/src-d/code-annotation/issues/149

screen shot 2018-02-21 at 17 24 48

The page is available by url `/dashboard`.

PR #156, opened 2018-02-22
Connect experiments page

smacker:

Based on:
https://github.com/src-d/code-annotation/pull/152
https://github.com/src-d/code-annotation/pull/154

`mark as finished` link is commented, we don't have API for it now.

there are +2 queries to DB for each experiment, we can optimize it later if we need (if we will have like thousands of experiments). For now, I decided to keep it more flexible not performant.

### DIFF
--- a/server/repository/assignments.go
+++ b/server/repository/assignments.go
@@ -24,8 +24,8 @@ const (
 	selectAssignmentsWhereExpPairSQL = `SELECT * FROM assignments WHERE experiment_id=$1 AND pair_id=$2`
 	updateAssignmentsSQL             = `UPDATE assignments SET answer='%v', duration=%v WHERE id=%v`
 	countPendingIDsSQL               = `SELECT count(id) FROM file_pairs WHERE experiment_id=$1 AND id NOT IN (SELECT pair_id FROM assignments WHERE experiment_id=$1 AND user_id=$2)`
-	countUserAssigmentsSQL           = `SELECT COUNT(*) FROM assignments WHERE experiment_id=$1 AND user_id=$1`
-	countCompleteUserAssigmentsSQL   = `SELECT COUNT(*) FROM assignments WHERE experiment_id=$1 AND user_id=$1 AND answer IS NOT null`
+	countUserAssigmentsSQL           = `SELECT COUNT(*) FROM assignments WHERE experiment_id=$1 AND user_id=$2`
+	countCompleteUserAssigmentsSQL   = `SELECT COUNT(*) FROM assignments WHERE experiment_id=$1 AND user_id=$2 AND answer IS NOT null`
 )
 
 // IsInitialized returns true if the assignments are initialized for the given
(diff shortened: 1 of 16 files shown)


server/repository/assignments.go
```diff
@@ -24,6 +24,8 @@ const (
 	selectAssignmentsWhereExpPairSQL = `SELECT * FROM assignments WHERE experiment_id=$1 AND pair_id=$2`
 	updateAssignmentsSQL             = `UPDATE assignments SET answer='%v', duration=%v WHERE id=%v`
 	countPendingIDsSQL               = `SELECT count(id) FROM file_pairs WHERE experiment_id=$1 AND id NOT IN (SELECT pair_id FROM assignments WHERE experiment_id=$1 AND user_id=$2)`
+	countUserAssigmentsSQL           = `SELECT COUNT(*) FROM assignments WHERE experiment_id=$1 AND user_id=$1`
+	countCompleteUserAssigmentsSQL   = `SELECT COUNT(*) FROM assignments WHERE experiment_id=$1 AND user_id=$1 AND answer IS NOT null`
```
carlosms: These two queries are using $1 for both arguments, user_id should be $2

server/repository/assignments.go
```diff
@@ -152,3 +154,27 @@ func (repo *Assignments) Update(assignmentID int, answer string, duration int) e
 
 	return err
 }
+
+// CountUserAssigments returns number of assigments in given experiment for the given user
+func (repo *Assignments) CountUserAssigments(experimentID, userID int) (int, error) {
```
carlosms: Typo in several places, Assigment -> assignment

carlosms (changes requested):
I found a small bug, and typo

server/repository/assignments.go
smacker: thank you so much! fixed

server/repository/assignments.go
smacker: thank you! I never spell it correctly

bzz:
@carlosms seems like all the feedback has been addressed now and it's ready for another pass.

bzz (approved):
LGTM

bzz:
@carlosms do you think there still is something left before merging this?

carlosms:
Right now there isn't a way to go to the dashboard. Upper left link takes you to `/exp/1/...` and the "dashboard" text in the upper right opens the dropdown, and in that drop down you can only click "review" and "logout".

I'm not sure if this problem should be solved in this PR, but it's here where I've noticed it.

ricardobaeta:
@carlosms Thank you for the notice. I would suggest that the "logo/upper-left-link" takes people to the dashboard. And the "dashboard" link would change its label to "more" or "options" with the current dropdown "review/log out".

smacker:
@carlosms it's solved in next PR. I kept dashboard unlinkable here on purpose because it has links to other experiments and those don't work yet. (I fixed it in next PR and dashboard is a default index page for logged in user)

carlosms (approved):
:+1:

PR #157, opened 2018-02-22
Fe multiple experiments support

smacker:

Based on:
https://github.com/src-d/code-annotation/pull/152
https://github.com/src-d/code-annotation/pull/154
https://github.com/src-d/code-annotation/pull/156

Now the list of experiments is default page when user login.
Annotation works fine for experiments with id > 1.

Review link still points on the first experiment though. It's possible to open review for other experiments changing url, but no link. Most probably the experiments list page should be modified and links should be there. After we can remove the one from top bar. But we don't have wireframes for that.

### DIFF
--- a/src/pages/Experiment.js
+++ b/src/pages/Experiment.js
@@ -11,7 +11,6 @@ import Diff from '../components/Experiment/Diff';
 import Selector from '../components/Experiment/Selector';
 import Actions from '../components/Experiment/Actions';
 import AdditionActions from '../components/Experiment/AdditionActions';
-import { experimentId } from '../state/experiment';
 import {
   markCurrent,
   ANSWER_SIMILAR,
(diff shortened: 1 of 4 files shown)


bzz:
It's quite hard to tell but I assume it's 108d583 that needs to be reviewed, would appreciate if you could you please rebase and verify this assumption.


Would you mind, in such cases when work is based on many other PRs listing the commits relevant for review if possible? That will speed up the reviews.

smacker:
@bzz commit with name `Fe multiple experiments support` should be reviewed (hash is changing after every rebase)

bzz (approved):
20ce24d SGTM

dpordomingo (approved):
LGTM, could you rebase and merge @smacker ?

Issue #160, opened 2018-02-23
Experiments management


smacker:
We need to allow to manage experiments by web ui:

 - [x] create endpoint to manage experiments POST #180
 - [x] create endpoints to manage experiments PUT #190
 - [x] create an endpoint to upload file pairs to experiment #188 
 - [x] create wireframes for UI
 - [x] create frontend for things above #195

carlosms:
Hey @ricardobaeta we are ready to start working on the new front-end code to manage experiments. The new functionality, available only to Reviewers, is:

- Upload a DB file with file pairs to be added to a single experiment
- Create a new experiment: Name and Description can be provided
- Update the Name or Description of an existing experiment

Please let us know if you prefer to provide wireframes beforehand, or if it's better if we build something simple and then iterate the design with your input.
I imagine it could be similar to the current dashboard, with buttons to 'edit' 'add file pairs', and 'create new experiment'

ricardobaeta:
Hey @carlosms The wireframes for this UI is in my to-do for today! Also, updated the description to include the wireframming.

ricardobaeta:
@bzz @dpordomingo @smacker 

Thanks for the help @carlosms !

Here's the wireframe https://balsamiq.cloud/sdjen2c/puva1qi

carlosms:
Thanks @ricardobaeta. Just to confirm, this wireframe shows the current dashboard with extra elements visible only to users with the right role, or should this be a separate page accessible from the menu?

ricardobaeta:
> Just to confirm, this wireframe shows the current dashboard with extra elements visible only to users with the right role, or should this be a separate page accessible from the menu?

Let's make the experience as collaborative as we can. ```visible only to users with the right role``` Persona disclosure :)

dpordomingo:
LGTM, but instead of «survey» I'd use «experiment»:

I'd also clarify in the upload section that the uploaded data will be **added** to the current one.

![image](https://user-images.githubusercontent.com/2437584/36980444-8cc476a8-208a-11e8-8ab5-8e7c13789ab1.png)

I say so because the button says «upload new file», what can be understood as replacing current data with new one. What about if the button says: «add more file pairs», for example...

ricardobaeta:
> but instead of «survey» I'd use «experiment»

The correct semantic expression is *survey*. Articulated in the past with @vmarkovtsev and @marnovo  

> and I'd clarify in the upload section that the uploaded data will be added to the current one.
[...]

PR #161, opened 2018-02-23
Import command multiple experiments support

smacker:
Partly implements https://github.com/src-d/code-annotation/issues/149

Looks like we don't really need to change any other commands.
There is no way to create new experiment through command line, but I think it's better to work on https://github.com/src-d/code-annotation/issues/160 than creating more commands.
These changes are necessary because import itself will be used in handler.


carlosms (approved):
LGTM

bzz (approved):
LGTM

PR #162, opened 2018-02-23
Add google analytics page view tracking

smacker:
partially solves: https://github.com/src-d/code-annotation/issues/140
(as described in the issue it tracks only urls)


bzz (approved):
LGTM

carlosms (changes requested):
Looks good to me.

I have one request: Add this new environment to the docs (will be done in #170)
And one question: should this be added to the helm charts somewhere? ping @dpordomingo

server/handler/static_files.go
```diff
@@ -10,18 +10,20 @@ import (
...
-	return &Static{dir, serverURL}
+func NewStatic(dir, serverURL, GaTrackingID string) *Static {
+	return &Static{dir, serverURL, GaTrackingID}
 }
 
 type options struct {
-	ServerURL   string      `json:"SERVER_URL"`
-	InitalState interface{} `json:"initialState"`
+	ServerURL    string      `json:"SERVER_URL"`
+	GaTrackingID string      `json:"GA_TRACKING_ID"`
```
dpordomingo, line 31: This value was not passed to the [marshalled options](https://github.com/src-d/code-annotation/pull/162/files#diff-33aa342cfbbafa0c5a364a50258cd956R50), so it does not get the frontend

cli/server/server.go
```diff
@@ -17,13 +17,14 @@ import (
...
-	DBConn      string `envconfig:"DB_CONNECTION" default:"sqlite:///var/code-annotation/internal.db"`
-	ExportsPath string `envconfig:"EXPORTS_PATH" default:"./exports"`
+	Env          string `envconfig:"ENV" default:"production"`
+	Host         string `envconfig:"HOST" default:"0.0.0.0"`
+	Port         int    `envconfig:"PORT" default:"8080"`
+	ServerURL    string `envconfig:"SERVER_URL"`
+	UIDomain     string `envconfig:"UI_DOMAIN"`
+	DBConn       string `envconfig:"DB_CONNECTION" default:"sqlite:///var/code-annotation/internal.db"`
+	ExportsPath  string `envconfig:"EXPORTS_PATH" default:"./exports"`
+	GaTrackingID string `envconfig:"GA_TRACKING_ID" required:"false"`
```
dpordomingo, line 27: We should define this value in the `helm-charts/values.yml`, inside the [`deployment` section](https://github.com/src-d/code-annotation/blob/master/helm-charts/code-annotation/values.yaml#L13). Since we do not know yet its production value, we could use an empty string; it will be replaced with the final `GA_TRACKING_ID` value once it has been provided by https://github.com/src-d/code-annotation/issues/169

The value above should be passed to the env at [helm-charts/templates/deployment.yaml](https://github.com/src-d/code-annotation/blob/master/helm-charts/code-annotation/templates/deployment.yaml#L28)

dpordomingo (changes requested):
LGTM, but there are some things to address

server/handler/static_files.go
smacker: fixed. Thanks a lot!

cli/server/server.go
smacker: added to helm chart too

carlosms (approved):

dpordomingo (approved):
LGTM,
can you rebase and merge @smacker ?

PR #164, opened 2018-02-26
Fix crash on not found

smacker:
In the test, there are no actions because the store doesn't use enhancer from redux-little-router.


carlosms (approved):

dpordomingo (approved):
LGTM
@smacker you can merge

PR #172, opened 2018-02-27
Fix DB lock bug in import code

carlosms:
While importing new file pairs SQLite to SQLite I was getting the error message "database is locked".
This commit fixes it closing the first connection instead of deferring it.


dpordomingo (approved):
LGTM

But we should avoid these kind of functions which doesn't fit in a window without scrolling, reuses variables inside it's scope and does different things. For example, I think this bug would be avoided if the existence of the Experiment would have been checked in a separated —and smaller— function.

bzz (approved):
good spot, LGTM

PR #188, opened 2018-03-01
Upload endpoint

smacker:
Implements endpoint for https://github.com/src-d/code-annotation/issues/160

PR #190, opened 2018-03-01
Experiment update endpoint

smacker:
Based on #188

Implements endpoint for #160


dpordomingo (approved):
LGTM

bzz:
May be that is something obvious, but from the description it is not very clear, in what use-case update of an experiment is going to happen.

bzz (approved):

smacker:
@bzz as usual button edit that allows to change name or description.

Issue #176, opened 2018-02-28
Show new lines, tabs, whitespaces


EgorBu:
Hi,
it may help to understand why diff behaves not so obvious in some cases (https://github.com/src-d/code-annotation/issues/174 + pair 29)

bzz:
@ricardobaeta we are talking about triggering an optional mode of the text editors, that shows non-printable characters, something like ![editor example](http://netbeans.dzone.com/sites/all/files/show-spaces-1.png)

Could you please think about it and provide a screenshot, that have the control element for this action placed somewhere in the annotator UI? I would imagine that a checkbox might be a good fit as it represents the state.

@smacker AFAIK we are using https://github.com/rtfpessoa/diff2html to display the diff - could you please provide some estimates on how much would would it be, to enable this behavior for the components displaying the text?

smacker:
@bzz maybe we shouldn't touch diff2html. According to the issue title we can just preprocess diffs, replacing characters.

smacker:
I changed the title of the issue removing `etc` from it. So it's feasible to implement now.
If we need to show any other characters please update issue or create a new one.

ricardobaeta:
@bzz @smacker I'm a bit lost here. Is this moving forward now? Wireframing I mean.

smacker:
@ricardobaeta I'm starting implementation now. It would be cool to have control somewhere to switch off/on this feature. But for diff itself, we can start with the simplest approach (don't change anything, just replace symbols).

smacker:
Example of spaces and new lines:
<img width="1200" alt="screen shot 2018-03-02 at 17 41 47" src="https://user-images.githubusercontent.com/406916/36910445-1348a308-1e41-11e8-8d5c-ce41d93b9f78.png">

Example of tabs to spaces (my branch vs staging):
<img width="1201" alt="screen shot 2018-03-02 at 17 40 49" src="https://user-images.githubusercontent.com/406916/36910480-24587f56-1e41-11e8-98ec-c2cf89a7a239.png">
<img width="1200" alt="screen shot 2018-03-02 at 17 40 59" src="https://user-images.githubusercontent.com/406916/36910483-286858fa-1e41-11e8-9252-23a69a181995.png">

I need a switcher to continue.

ricardobaeta:
@smacker I'm working on the wireframe, thanks for the notice :)

ricardobaeta:
@smacker @bzz @dpordomingo @carlosms

[...]

Issue #112, opened 2018-02-15
Documents are truncated


r0mainK:
In the case of the second file below there are not many changes, but the file seems truncated from lines 17 to 89/99 so it's hard to tell whether the file is similar (e.g. if the file is 1000 lines long) or different (e.g. if it's 100). 

<img width="1207" alt="screen shot 2018-02-15 at 10 06 32 1" src="https://user-images.githubusercontent.com/32878976/36248584-5009afb8-1238-11e8-8a31-84fe79282dc6.png">

bzz:
Thank you for the feedback!

I'm not sure I understand 
> the file seems truncated from lines 17 to 89/99

could you please elaborate on this a bit?

r0mainK:
Sure, well on the left of the code you can see the line number start at 17, but I cannot scroll up to see the first 16 lines. Also since on lines 89 and 99 there is an `@Override` I guess there is a function below but I canno't scroll down to see it. That's a problem because even if the files are the same below and under, the length of the document will impact on whether the file is categorized as different or similar.

bzz:
Got it, thank you for explaining @r0mainK !

@smacker @carlosms how much work do you guys think it would be to switch from a diff rending \w `@@ -19,106 +15,106 @@` to full file only with the color highlight for diffs?

smacker:
Most probably we just need to change `Context` value of `UnifiedDiff`. Now it's 3. If we put a really big number there it should work. Not the most pretty way to solve the issue, but the easiest one.
The only problem - we store diffs in DB, so the change won't appear on staging after deployment.

EgorBu:
Hi,
another good thing that may help - possibility to switch on/off highlights and diff so we can see full contents of files.
In some cases, highlights/diff are misleading (ex: https://github.com/src-d/code-annotation/issues/174 & #113 ).
Why it may help - if a diff is wrong you may need to scroll a lot to see the beginning of "new" part of the code - it's quite inconvenient.

EgorBu:
Another example why full content can be useful is pair 325, it's hard to understand what was modified

smacker:
I return issue in backlog. Some improvements were done in #194 but it doesn't solve the issue completely. Let's see how it works and decide what to do later.

bzz:
As soon as it has been addressed and there is no more feedback, closing it for now.

Please feel free to reopen it if this would still be an issue at any time later on.

PR #194, opened 2018-03-02
Make diff service and increase context

smacker:
Improvements for https://github.com/src-d/code-annotation/issues/112
Moving generation of diff to the handler and increasing context. In real life discussion, guys said most probably 6 lines will be alright. Later with diff generation in the handler, we will be able to increase it further or add configuration in UI or do any other improvements easier if we need.

Also, it allows to preprocess diff on backend for https://github.com/src-d/code-annotation/issues/176
(rendering diff is already expensive, adding preprocessing on FE would degrade user experience further)


bzz (approved):
LGTM

👍 for tests and flexibility of diff generation

dpordomingo (approved):
So now diffs are going to be calculated every time in backend on every request... ok... LGTM

(let's add some cache in the future if performance is impacted)

dpordomingo:
you can merge @smacker

smacker:
I didn't add cache now because diff generation is actually very fast. If we ever meet performance problems it will be easy to add it.

PR #195, opened 2018-03-02
Manage Experiments from the front

carlosms:
Part of #160, this PR will add a new route to create & edit experiments, and a way to upload a new SQLite DB with file pairs.

### DIFF
--- a/src/components/Experiments/ExperimentEditModal.js
+++ b/src/components/Experiments/ExperimentEditModal.js
@@ -1,0 +1,182 @@
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+
+import {
+  Button,
+  Modal,
+  FormGroup,
+  FormControl,
+  Glyphicon,
+  Row,
+  Col,
+} from 'react-bootstrap';
+
+import {
+  create as experimentCreate,
+  update as experimentUpdate,
+  uploadFilePairs,
+  UPLOAD_RES_SUCCESS,
+  UPLOAD_RES_FAILURE,
+} from '../../state/experiments';
+
+import './ExperimentEditModal.less';
+
+class ExperimentEditModal extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = this.buildState(this.props);
+
+    this.handleNameChange = this.handleNameChange.bind(this);
+    this.handleDescriptionChange = this.handleDescriptionChange.bind(this);
+    this.handleSave = this.handleSave.bind(this);
+    this.handleUpload = this.handleUpload.bind(this);
+  }
+
+  buildState(props) {
+    const exp = props.editModalExperiment;
+
+    if (props.action === 'edit' && exp !== undefined) {
+      return { nameVal: exp.name, descriptionVal: exp.description };
+    }
+
+    return { nameVal: '', descriptionVal: '' };
+  }
+
+  handleNameChange(e) {
+    this.setState({ nameVal: e.target.value });
+  }
+
+  handleDescriptionChange(e) {
+    this.setState({ descriptionVal: e.target.value });
+  }
+
+  handleSave() {
+    const create = this.props.action === 'create';
+
+    if (create) {
+      return this.props
+        .experimentCreate(this.state.nameVal, this.state.descriptionVal)
+        .then(expId => {
+          this.props.onOpenEdit(expId);
+        })
+        .catch(() => {});
+    }
+
+    return this.props.experimentUpdate(
+      this.props.editModalExperiment.id,
+      this.state.nameVal,
+      this.state.descriptionVal
+    );
+  }
+
+  handleUpload(e) {
+    this.props.uploadFilePairs(
+      this.props.editModalExperiment.id,
+      e.target.files[0]
+    );
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.editModalExperiment !== this.props.editModalExperiment) {
+      this.setState(this.buildState(nextProps));
+    }
+  }
+
+  render() {
+    const {
+      createInProgress,
+      updateInProgress,
+      uploadInProgress,
+      uploadResult,
+    } = this.props.experiments;
+
+    const create = this.props.action === 'create';
+
+    const title = create ? 'Create Experiment' : 'Edit Experiment';
+    const btn = create ? 'Create' : 'Save';
+    const btnDisabled =
+      createInProgress || updateInProgress || this.state.nameVal.trim() === '';
+
+    const uploadDisabled = create || uploadInProgress;
+
+    let uploadIcon;
+    let uploadClass;
+    if (uploadInProgress) {
+      uploadIcon = 'floppy-open';
+      uploadClass = 'upload-floppy upload-floppy-progress';
+    } else if (uploadResult === UPLOAD_RES_SUCCESS) {
+      uploadIcon = 'floppy-saved';
+      uploadClass = 'upload-floppy upload-floppy-success';
+    } else if (uploadResult === UPLOAD_RES_FAILURE) {
+      uploadIcon = 'floppy-remove';
+      uploadClass = 'upload-floppy upload-floppy-failure';
+    } else {
+      uploadIcon = 'floppy-disk';
+      uploadClass = 'upload-floppy';
+    }
+
+    return (
+      <Modal show={this.props.show} onHide={this.props.onHide}>
+        <Modal.Header closeButton>
+          <Modal.Title>{title}</Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
+          <FormGroup controlId="nameInput">
+            <FormControl
+              type="text"
+              value={this.state.nameVal}
+              placeholder="Experiment name"
+              onChange={this.handleNameChange}
+            />
+          </FormGroup>
+          <FormGroup controlId="descInput">
+            <FormControl
+              type="text"
+              value={this.state.descriptionVal}
+              placeholder="Experiment description"
+              onChange={this.handleDescriptionChange}
+            />
+          </FormGroup>
+          <hr />
+          <Row>
+            <Col xs={7}>
+              <p>
+                <Glyphicon className={uploadClass} glyph={uploadIcon} />
+                <b> Experiment Dataset</b> SQLite DB file
+              </p>
+            </Col>
+            <Col xs={5} className="text-right">
+              <FormGroup controlId="uploadTest">
+                <FormControl
+                  type="file"
+                  onChange={this.handleUpload}
+                  disabled={uploadDisabled}
+                />
+              </FormGroup>
+            </Col>
+          </Row>
+        </Modal.Body>
+        <Modal.Footer>
+          <Button
+            bsStyle="primary"
+            disabled={btnDisabled}
+            onClick={this.handleSave}
+          >
+            {btn}
+          </Button>
+        </Modal.Footer>
+      </Modal>
+    );
+  }
+}
+
+const mapStateToProps = state => ({
+  experiments: state.experiments,
+});
+
+export default connect(mapStateToProps, {
+  experimentCreate,
+  experimentUpdate,
+  uploadFilePairs,
+})(ExperimentEditModal);
--- a/src/components/ExperimentsList.js
+++ b/src/components/ExperimentsList.js
@@ -1,58 +1,82 @@
-import React from 'react';
-import { Table, Button } from 'react-bootstrap';
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import { Table, Button, Glyphicon } from 'react-bootstrap';
 import { makeUrl } from '../state/routes';
 import './ExperimentsList.less';
 
-function buttonText(progress) {
-  switch (progress) {
-    case 0:
-      return 'Begin';
-    case 100:
-      return 'Finished';
-    default:
-      return 'Continue';
+class ExperimentsList extends Component {
+  handleOpenEdit(expId) {
+    return () => this.props.onOpenEdit(expId);
+  }
+
+  buttonText(progress) {
+    switch (progress) {
+      case 0:
+        return 'Begin';
+      case 100:
+        return 'Finished';
+      default:
+        return 'Continue';
+    }
+  }
+
+  render() {
+    return (
+      <Table responsive className="experiments-list">
+        <tbody>
+          {this.props.experiments.map(exp => {
+            // const started = exp.progress === 0;
+            const finished = exp.progress === 100;
+            return (
+              <tr
+                key={exp.id}
+                className={`experiments-list__row${
+                  finished ? ' finished' : ''
+                }`}
+              >
+                <td className="experiments-list__name">{exp.name}</td>
+                {this.props.user.role === 'requester' && (
+                  <td className="experiments-list__edit">
+                    <Button
+                      bsStyle="link"
+                      onClick={this.handleOpenEdit(exp.id)}
+                    >
+                      <Glyphicon glyph="edit" />
+                    </Button>
+                  </td>
+                )}
+                <td className="experiments-list__progress">
+                  {Math.round(exp.progress)}%
+                </td>
+                <td className="experiments-list__additional-actions">
+                  {/*
+                    {!finished && !started ? (
+                        <a href="/">mark as finished</a>
+                      ) : null}
+                  */}
+                </td>
+                <td className="experiments-list__actions">
+                  <Button
+                    bsStyle="primary"
+                    bsSize="xsmall"
+                    disabled={finished}
+                    href={makeUrl('experiment', { experiment: exp.id })}
+                  >
+                    {this.buttonText(exp.progress)}
+                  </Button>
+                </td>
+              </tr>
+            );
+          })}
+        </tbody>
+      </Table>
+    );
   }
 }
 
-function ExperimentsList({ experiments }) {
-  return (
-    <Table responsive className="experiments-list">
-      <tbody>
-        {experiments.map(exp => {
-          // const started = exp.progress === 0;
-          const finished = exp.progress === 100;
-          return (
-            <tr
-              key={exp.id}
-              className={`experiments-list__row${finished ? ' finished' : ''}`}
-            >
-              <td className="experiments-list__name">{exp.name}</td>
-              <td className="experiments-list__progress">
-                {Math.round(exp.progress)}%
-              </td>
-              <td className="experiments-list__additional-actions">
-                {/*
-                  {!finished && !started ? (
-                      <a href="/">mark as finished</a>
-                    ) : null}
-                */}
-              </td>
-              <td className="experiments-list__actions">
-                <Button
-                  bsStyle="primary"
-                  bsSize="xsmall"
-                  disabled={finished}
-                  href={makeUrl('experiment', { experiment: exp.id })}
-                >
-                  {buttonText(exp.progress)}
-                </Button>
-              </td>
-            </tr>
-          );
-        })}
-      </tbody>
-    </Table>
-  );
-}
+const mapStateToProps = state => ({
+  user: state.user,
+  experiments: state.experiments.list,
+});
 
-export default ExperimentsList;
+export default connect(mapStateToProps)(ExperimentsList);
--- a/src/state/__snapshots__/experiments.test.js.snap
+++ b/src/state/__snapshots__/experiments.test.js.snap
@@ -1,31 +1,80 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`experiments/reducer LOAD 1`] = `
+exports[`experiments/reducer CREATE 1`] = `
 Object {
+  "createInProgress": true,
   "error": null,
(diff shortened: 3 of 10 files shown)


src/pages/ExperimentsAdmin.js
```diff
@@ -0,0 +1,96 @@
...
+            </Button>
+          </Col>
+        </Row>
+      </React.Fragment>
+    );
+  }
+}
+
+const mapStateToProps = state => ({
+  user: state.user,
```
smacker: looks like you don't use it anywhere

src/state/experiments.js
```diff
@@ -6,6 +6,7 @@ import { add as addErrors } from './errors';
 export const initialState = {
   loading: true,
   error: null,
+  createOnProgress: false,
```
smacker: shouldn't it be `in progress`?

src/pages/ExperimentsAdmin.js
```diff
@@ -0,0 +1,96 @@
...
+import Loader from '../components/Loader';
+import ExperimentsAdminList from '../components/ExperimentsAdminList';
+import {
+  load as experimentsLoad,
+  create as experimentCreate,
+} from '../state/experiments';
+
+class ExperimentsAdmin extends Component {
+  componentDidMount() {
+    this.props.load();
```
smacker: no need to fix it, just to let you know. Here we have kind of "bug" here. Because user will come here from another page in most of the cases we will have experiments in the store already. So a user will see the list of experiments and in few milliseconds it will be replaced by a loader and then by list again.

src/state/experiments.js
```diff
@@ -61,3 +87,17 @@ export const load = () => dispatch => {
       dispatch({ type: LOAD_ERROR, error: e });
     });
 };
+
+export const create = (name, description) => dispatch => {
+  dispatch({ type: CREATE });
+  return api
+    .createExperiment(name, description)
+    .then(() => {
+      dispatch({ type: CREATE_SUCCESS });
```
smacker: no need to fix. Just let you know, an optimistic update would fit nicely here.

src/pages/ExperimentsAdmin.js
carlosms: This was actually copied from Experiments.js.
I guess one way to handle this would be to have an expiration date for the data, and then the function could load only if the list is empty, or the expiration date passed... But for now in my opinion a forced load is not that inconvenient for the user experience.

carlosms:
This is how it looks, based on the wireframes provided by @ricardobaeta here https://github.com/src-d/code-annotation/issues/160#issuecomment-369960677, but keeping the name 'experiment' for consistency with the rest of the app.

peek 2018-03-09 19-36

dpordomingo:
Conflicts appeared :(

src/state/experiments.js
```diff
@@ -3,9 +3,20 @@ import { add as addErrors } from './errors';
...
 export const initialState = {
   loading: true,
   error: null,
+  createInProgress: false,
+  updateInProgress: false,
+  uploadInProgress: false,
+  uploadResult: UPLOAD_RES_NONE,
+
+  editModalShow: false,
+  editModalExpID: null,
```
smacker: could you please keep names consistent with all other js files?
In js in general and in code-annotation frontend camelCase is a standard.

src/state/experiments.js
```diff
@@ -14,11 +25,27 @@ export const LOAD = 'ca/experiments/LOAD';
...
+export const CREATE = 'ca/experiments/CREATE';
+export const CREATE_SUCCESS = 'ca/experiments/CREATE_SUCCESS';
+export const CREATE_ERROR = 'ca/experiments/CREATE_ERROR';
+export const UPDATE = 'ca/experiments/UPDATE';
+export const UPDATE_SUCCESS = 'ca/experiments/UPDATE_SUCCESS';
+export const UPDATE_ERROR = 'ca/experiments/UPDATE_ERROR';
+export const UPLOAD = 'ca/experiments/UPLOAD';
+export const UPLOAD_SUCCESS = 'ca/experiments/UPLOAD_SUCCESS';
+export const UPLOAD_ERROR = 'ca/experiments/UPLOAD_ERROR';
+export const OPEN_EDIT_MODAL = 'ca/experiments/OPEN_EDIT_MODAL';
```
smacker: usually, it's a good idea to keep store agnostic to presentation level.

src/components/Experiments/ExperimentEditModal.js
```diff
@@ -0,0 +1,195 @@
...
+
+  handleSave() {
+    const create = this.props.editModalExperiment === undefined;
+
+    if (create) {
+      this.props.experimentCreate(
+        this.state.nameVal,
+        this.state.descriptionVal
+      );
+    } else {
```
smacker: please follow `early return` rule as all other js code.

src/components/Experiments/ExperimentEditModal.js
```diff
@@ -0,0 +1,195 @@
...
+
+  handleUpload(e) {
+    this.props.uploadFilePairs(
+      this.props.editModalExperiment.id,
+      e.target.files[0]
+    );
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const exp = nextProps.editModalExperiment;
```
smacker: I see this code 2 times. The copy-past itself is not a problem. But component depends on both constructor and componentWillReceiveProps. So any the change must be done in both places. It would be much better to extract this code to function which returns state. And do `this.state = fn(props)` in constructor and `this.setState(fn(nextProps))` here.

src/components/Experiments/ExperimentEditModal.js
```diff
@@ -0,0 +1,195 @@
...
+    if (exp !== undefined) {
```
smacker: the check should compare nextProps with current props. If you need more info, come over and I can explain it verbally.

src/components/Experiments/ExperimentEditModal.js
```diff
@@ -0,0 +1,195 @@
...
+    const title = create ? 'Create Experiment' : 'Edit Experiment';
+    const btn = create ? 'Create' : 'Save';
+    const btnDisabled =
+      createInProgress || updateInProgress || this.state.nameVal.trim() === '';
+
+    const uploadDisabled = create || uploadInProgress;
+
+    let uploadIcon;
+    let uploadClass;
+    if (uploadInProgress) {
```
smacker, line 105: looks like `switch` will be better fit here

src/components/Experiments/ExperimentEditModal.js
```diff
@@ -0,0 +1,195 @@
...
+    } else if (uploadResult === UPLOAD_RES_FAILURE) {
+      uploadIcon = 'floppy-remove';
+      uploadClass = 'upload-floppy upload-floppy-failure';
+    } else {
+      uploadIcon = 'floppy-disk';
+      uploadClass = 'upload-floppy';
+    }
+
+    return (
+      <Modal show={editModalShow} onHide={this.handleClose}>
```
smacker: I would recommend controlling `show` outside of the modal component. In general it's better to do `editModalShow  && <Modal>`. Otherwise it might not render virtual/real dom if the underlying component is written well, but all life cycle methods would be executed on every parent update.
Not very important in this particular case though.

src/state/experiments.js
```diff
@@ -61,3 +155,50 @@ export const load = () => dispatch => {
...
+};
+
+export const create = (name, description) => dispatch => {
+  dispatch({ type: CREATE });
+  return api
+    .createExperiment(name, description)
+    .then(res => {
+      dispatch({ type: CREATE_SUCCESS });
+      dispatch(load());
+      dispatch(showEditModal(true, res.id));
```
smacker: here is a good example why it's better to keep store view agnostic. `create` action can be called from anywhere. But it will set the modal show to true. It is super confusing and can cause weird bugs.

smacker (changes requested):
some comments.
also tests are missed.

src/state/experiments.js
carlosms: I must be missing something, sorry, can you paste me the exact change you have in mind?

src/state/experiments.js
carlosms: To be honest I didn't like that much the previous code, but I didn't want to spend much more time refactoring it. I've tried to improve it in https://github.com/src-d/code-annotation/pull/195/commits/ce357ffaf5ea919e448cbcd238e52032ed6373cd, moving the modal state to Experiments.js

src/components/Experiments/ExperimentEditModal.js
carlosms: Changed in https://github.com/src-d/code-annotation/pull/195/commits/0fb33e0f998609b4e0e9304e7576194fb695b0ad

Can you please point me to a guideline (like the airbnb one we use in eslint.rb) where this rule is explained?

src/state/experiments.js
smacker: `editModalExpID` -> editModalExpId`. Just because it's `Id` not `ID` everywhere.

src/components/Experiments/ExperimentEditModal.js
carlosms: Makes sense, it's an easy improvement. Changed in https://github.com/src-d/code-annotation/pull/195/commits/dd9e4fb3943b1f65200e19dec7b9e6ce580973d7

src/components/Experiments/ExperimentEditModal.js
carlosms: Changed in https://github.com/src-d/code-annotation/pull/195/commits/dd9e4fb3943b1f65200e19dec7b9e6ce580973d7

src/components/Experiments/ExperimentEditModal.js
carlosms: Are you sure? The chained if's condition use two different variables, I don't think it will be cleaner with a switch.

ricardobaeta:
Thank you @carlosms :) It looks great!

src/components/Experiments/ExperimentEditModal.js
carlosms: Makes sense. I decided to leave it as a property of the Modal component to avoid introducing more and more changes. Let me know if you think this should be changed considering the last changes.

src/state/experiments.js
carlosms: Changed in https://github.com/src-d/code-annotation/pull/195/commits/ce357ffaf5ea919e448cbcd238e52032ed6373cd

carlosms:
@dpordomingo I rebased and there are no more conflicts (I was avoiding it to not loose the commit hashes).

@smacker feel free to give it another pass. I was delaying making the tests to wait for change requests, I'll upload them shortly.

src/components/Experiments/ExperimentEditModal.js
smacker: the rule usually you look how the current code is written and do the same.

there is a rule about no-else return in airbnb:
https://github.com/airbnb/javascript#blocks--no-else-return

no **always** early return because sometimes it makes sense to not return early (though it was there before)

you can read more why people do that by simple googling `early return`. It's very popular pattern in javascript. Java style of `else` everywhere is much less common.

src/state/experiments.js
```diff
@@ -14,11 +22,26 @@ export const LOAD = 'ca/experiments/LOAD';
 export const LOAD_SUCCESS = 'ca/experiments/LOAD_SUCCESS';
 export const LOAD_ERROR = 'ca/experiments/LOAD_ERROR';
 export const SET = 'ca/experiments/SET';
+export const CREATE = 'ca/experiments/CREATE';
```
smacker, line 25: it's not really wrong, but kind of weird. You can see how it's done in redux usually in official examples: https://github.com/reactjs/redux/tree/master/examples/todomvc/src

src/components/Experiments/ExperimentEditModal.js
```diff
@@ -0,0 +1,183 @@
...
+    const create = this.props.action === 'create';
+
+    if (create) {
+      this.props
+        .experimentCreate(this.state.nameVal, this.state.descriptionVal)
+        .then(expId => {
+          this.props.onOpenEdit(expId);
+        });
+
+      return;
```
smacker: not required here, but usually it's a good idea when you have a promise to return it. For example, you would need it for tests.

src/components/Experiments/ExperimentEditModal.js
```diff
@@ -0,0 +1,183 @@
...
+    this.setState({ descriptionVal: e.target.value });
+  }
+
+  handleSave() {
+    const create = this.props.action === 'create';
+
+    if (create) {
+      this.props
+        .experimentCreate(this.state.nameVal, this.state.descriptionVal)
+        .then(expId => {
```
smacker: I usually put such logic in dispatchToProps, to keep code of the component abstract. Not a big problem though.

src/state/__snapshots__/experiments.test.js.snap
```diff
@@ -1,37 +1,200 @@
...
+  "uploadInProgress": false,
+  "uploadResult": "ca/experiments/UPLOAD_RES_NONE",
+}
+`;
+
+exports[`experiments/reducer CREATE_ERROR 1`] = `
+Object {
+  "createInProgress": false,
+  "error": "test error",
+  "initialState": Object {
```
smacker: it's incorrect.

src/state/__snapshots__/experiments.test.js.snap
```diff
@@ -1,37 +1,200 @@
...
+    "uploadInProgress": false,
+    "uploadResult": "ca/experiments/UPLOAD_RES_NONE",
+  },
+}
+`;
+
+exports[`experiments/reducer CREATE_SUCCESS 1`] = `
+Object {
+  "createInProgress": false,
+  "initialState": Object {
```
smacker: it's incorrect.

src/state/__snapshots__/experiments.test.js.snap
```diff
@@ -1,37 +1,200 @@
...
+  "updateInProgress": true,
+  "uploadInProgress": false,
+  "uploadResult": "ca/experiments/UPLOAD_RES_NONE",
+}
+`;
+
+exports[`experiments/reducer UPDATE_ERROR 1`] = `
+Object {
+  "error": "test error",
+  "initialState": Object {
```
smacker: it's incorrect.

src/state/__snapshots__/experiments.test.js.snap
```diff
@@ -1,37 +1,200 @@
...
+    "uploadInProgress": false,
+    "uploadResult": "ca/experiments/UPLOAD_RES_NONE",
+  },
+  "updateInProgress": false,
+}
+`;
+
+exports[`experiments/reducer UPDATE_SUCCESS 1`] = `
+Object {
+  "initialState": Object {
```
smacker: it's incorrect.

src/components/Experiments/ExperimentEditModal.js
carlosms: Ok, thanks for clarifying. I'm aware of the rules you mention, but I just didn't see them applying here and wanted to check if you were talking about something else. From my point of view the code was:
```
if (condition)
    somehting()
else
    other()
end

// Posibly some more code
```

Now I understand your point of view, as there was no more code after the else.

src/components/Experiments/ExperimentEditModal.js
smacker: in case of error here will be undefined and `this.props.onOpenEdit` will fail

smacker (changes requested):
- tests are incorrect
-  found 1 bug
- few suggestions (not required, up to you)

src/state/__snapshots__/experiments.test.js.snap
carlosms: Yeah, I missed the spread operator while c&p and of course the tests were not failing :smile:

src/components/Experiments/ExperimentEditModal.js
carlosms: Done :+1:

src/state/experiments.test.js
```diff
@@ -110,4 +220,144 @@ describe('experiments/actions', () => {
...
+      });
+    });
+
+    it('error', () => {
+      const store = mockStore({ experiments: initialState });
+
+      const errText = 'some error';
+      fetch.mockReject(errText);
+
+      store.dispatch(create('new_name', 'new_desc')).catch(() => {
```
smacker: problem with this test: if the action doesn't throw any error, test still pass.

src/components/Experiments/ExperimentEditModal.js
```diff
@@ -0,0 +1,182 @@
...
+          <hr />
+          <Row>
+            <Col xs={7}>
+              <p>
+                <Glyphicon className={uploadClass} glyph={uploadIcon} />
+                <b> Experiment Dataset</b> SQLite DB file
+              </p>
+            </Col>
+            <Col xs={5} className="text-right">
+              <FormGroup controlId="uploadTest">
```
dpordomingo, line 150: If `uploadDisabled` is true because it is being created a new experiment, the upload button should not appear at all.

src/components/ExperimentsList.js
```diff
@@ -1,58 +1,82 @@
...
+                    >
+                      <Glyphicon glyph="edit" />
+                    </Button>
+                  </td>
+                )}
+                <td className="experiments-list__progress">
+                  {Math.round(exp.progress)}%
+                </td>
+                <td className="experiments-list__additional-actions">
+                  {/*
```
dpordomingo, line 52: leftover?

src/components/Experiments/ExperimentEditModal.js
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  handleSave() {
+    const create = this.props.action === 'create';
+
+    if (create) {
+      return this.props
+        .experimentCreate(this.state.nameVal, this.state.descriptionVal)
+        .then(expId => {
+          this.props.onOpenEdit(expId);
```
dpordomingo, line 61: It is kind of weird that after creating an experiment, the same modal is opened again without giving any special feedback to the user.

After creating a new experiment, I wouldn't open the modal automatically (as in edition-mode), so the user should open it manually -> she/he will be then aware of the edit-state.

src/components/Experiments/ExperimentEditModal.js
```diff
@@ -0,0 +1,182 @@
...
+        })
+        .catch(() => {});
+    }
+
+    return this.props.experimentUpdate(
+      this.props.editModalExperiment.id,
+      this.state.nameVal,
+      this.state.descriptionVal
+    );
```
dpordomingo, line 70: The save button should not open the same modal again without clear feedback.
- If the update succeeds, it should be said so, and the modal should be closed
- if the update fails, it should be shown an error

src/components/Experiments/ExperimentEditModal.js
```diff
@@ -0,0 +1,182 @@
...
+              <p>
+                <Glyphicon className={uploadClass} glyph={uploadIcon} />
+                <b> Experiment Dataset</b> SQLite DB file
+              </p>
+            </Col>
+            <Col xs={5} className="text-right">
+              <FormGroup controlId="uploadTest">
+                <FormControl
+                  type="file"
+                  onChange={this.handleUpload}
```
dpordomingo, line 153: I would not push/save the new DB till the user press "save".

dpordomingo (approved):
Wow!!! it worked perfectly! Many many thanks!
I only detected some (imho) UX problems that should be confirmed by @ricardobaeta  in a next iteration.

src/components/Experiments/ExperimentEditModal.js
dpordomingo: and `if` &rarr; `else switch` would be clearer, but I'd not block this wonderful work for this cosmetic suggestion :wink: 
```js
if (uploadInProgress) { /* something */ }
else switch (uploadResult) {
    case UPLOAD_RES_SUCCESS: /* something */ break;
    case UPLOAD_RES_FAILURE: /* something */ break;
    default: /* something */
}
```

dpordomingo:
@carlosms you did a great job, many thanks.

Since most important requirements were already addressed, we can merge, so let's proceed as it follows:
- you squash/rebase and whatever you think it should be done,
- then you merge,
- then I'll open an issue with mi UX concerns so @ricardobaeta will be able to address them,
- and I'll open a new issue with the issue in one testcase [as it was spotted](https://github.com/src-d/code-annotation/pull/195#pullrequestreview-103773783) by @smacker

src/components/ExperimentsList.js
carlosms: It's in master, I just changed the indentation :man_shrugging:

src/state/experiments.test.js
carlosms: Fixed: https://github.com/src-d/code-annotation/pull/195/commits/435f0865210bb6fdd571cfc5cb4d75670bb5f25d

smacker (approved):

PR #206, opened 2018-03-06
Use UAST columns from FilePairs table

### LINKED ISSUE #197: Receive and store serialized ProtoBuf \w UAST as part of the input
Right now, the [schema of sqlite input](https://github.com/src-d/code-annotation/blob/master/cli/import/examples/example.sql#L7) includes content of the 2 files and a git metadata.

After discussion \w @vmarkovtsev we determined the need to (optionally) add serialized UASTs of both files to that, which are going to be sent to the reviewer REST API for feature extraction.

This is crucial in order to enable upcoming function-level annotations.

### LINKED ISSUE #222: Blob is not supported in PostgreSQL
Since Blob is not supported by PostgreSQL as spotted by https://github.com/src-d/code-annotation/pull/206#pullrequestreview-102665598 its support was dropped by that https://github.com/src-d/code-annotation/pull/206

We should restore the PostgreSQL support and then revert https://github.com/src-d/code-annotation/pull/206/commits/94d41ab24c655a5f4b85c6960de882fef6ea730e if needed

smacker:
Fixes #197
Fixes #222

Blocks https://github.com/src-d/code-annotation/pull/221


server/repository/file_pairs.go
```diff
@@ -24,10 +24,10 @@ func (repo *FilePairs) getWithQuery(queryRow scannable) (*model.FilePair, error)
 
 	err := queryRow.Scan(&pair.ID,
 		&pair.Left.BlobID, &pair.Left.RepositoryID, &pair.Left.CommitHash,
-		&pair.Left.Path, &pair.Left.Content, &pair.Left.Hash,
+		&pair.Left.Path, &pair.Left.Content, &pair.Left.UAST, &pair.Left.Hash,
```
dpordomingo: In some cases, UAST can be really big, and it is not needed for printing FilePairs, so it could be removed from the query.
Let's keep it in mind for the future in case we notice that getting the next file pair is delaying too much

dpordomingo (approved):
LGTM

dpordomingo (changes requested):
I'm leaving this _red cross_ here to avoid merging this PR till we have considered/solved/scheduled migration

update: &rarr; https://github.com/src-d/code-annotation/issues/214

cli/import/examples/example.sql
```diff
@@ -4,8 +4,8 @@ Sample DB for the import command. You can create an sqlite DB with the command:
 $ sqlite3 example.db < example.sql
 */
 CREATE TABLE files (
-		blob_id_a TEXT, repository_id_a TEXT, commit_hash_a TEXT, path_a TEXT, content_a TEXT,
-		blob_id_b TEXT, repository_id_b TEXT, commit_hash_b TEXT, path_b TEXT, content_b TEXT,
+		blob_id_a TEXT, repository_id_a TEXT, commit_hash_a TEXT, path_a TEXT, content_a TEXT, uast_a BLOB,
+		blob_id_b TEXT, repository_id_b TEXT, commit_hash_b TEXT, path_b TEXT, content_b TEXT, uast_b BLOB,
```
carlosms, line 8: Type blob is not supported in postgres

carlosms (changes requested):

cli/import/examples/example.sql
dpordomingo: Since this is an internal tool, and (internally) we're not using PostgreSQL but SQLite, I'd not block this PR because of this problem with PostgreSQL.
If Blobs cannot be easily solved for PostgreSQL, I'd merge as it is, and create a new issue to work on it when facing [`v1`](https://github.com/src-d/code-annotation/labels/v1)

dpordomingo (commented):
Considering PostgreSQL issue pointed out by https://github.com/src-d/code-annotation/pull/206#pullrequestreview-102665598 if we decide to merge without providing support to PostgreSQL, we should remove from our docs the mentions to the support of PostgreSQL.

On the other hand, if the PostgreSQL issue is solved, we can disregard this comment

dpordomingo:
the plans for the migration were defined by https://github.com/src-d/code-annotation/issues/214

cli/import/examples/example.sql
bzz: @carlosms nice catch! 

@carlosms @smacker  Is there anything we could use instead that would allow to keep PostgreSQL compatibility?

carlosms: We could do something similar to this:
https://github.com/src-d/code-annotation/blob/master/server/dbutil/dbutil.go#L38

And use [`bytea` type for postgres](https://www.postgresql.org/docs/current/static/datatype-binary.html). We'd have to check if it works different or has any other side effect.

dpordomingo: If it requires extra work, I'd not block the merge as explained at https://github.com/src-d/code-annotation/pull/206#pullrequestreview-103163429

dpordomingo:
I rebased and fixed a bug in the original branch with 04f34a0

@carlosms @bzz could you review this PR considering https://github.com/src-d/code-annotation/pull/206#pullrequestreview-103163429 ?
If you don't strongly disagree with postponing the PostgreSQL support, we should merge in order to go to Prod with this new feature as soon as possible.

cli/import/examples/example.sql
bzz: Nice! Thank you for pointing it out @carlosms 

So, if that (and that's how it seems from example above) would require some custom code, to change the query at the runtime by replacing the type, right?

README.md
```diff
@@ -64,7 +64,7 @@ Download the binary from [releases](https://github.com/src-d/code-annotation/rel
 
 The pieces of code to be labeled are called _file pairs_. They must be provided via an [SQLite](https://sqlite.org/) database. The database **must follow the expected schema**, please [follow this link](./cli/import/examples/example.sql) to see an example.
 
-The `import` command will use those file pairs to create a new [SQLite](https://sqlite.org/) or [PostgreSQL](https://www.postgresql.org/) database that will be used internally by the Annotation Tool. The destination database does not need to be empty, new imported file pairs can be added to previous imports.
```
bzz: There is a discussion on [postgresql support](https://github.com/src-d/code-annotation/pull/206#discussion_r175986723) but at any rate, even if this PR breaks it - removing existing Posgresql support in this PR does not seem like a good idea to me.

If merging fast of this work is required, I would propose create a new issue for adding the support of switching a type on top of this work, and then implementing it asap after this is merged, rather then having both - adding new feature and removing an old one in the same PR.

bzz (changes requested):

cli/import/examples/example.sql
carlosms: That's right. For the `<INCREMENT_TYPE>` the substitution is done [in this method](https://github.com/src-d/code-annotation/blob/master/server/dbutil/dbutil.go#L142).

smacker:
Thanks a lot for fixing bug with an order of columns. I must be missed it.

Though I would disagree with the last commit. Same thoughts as of @bzz above. If we need to merge this PR now, just merge it as it is with BLOB type column without removing code for postgres altogether. And later we can fix it using bytea type.

README.md
dpordomingo: Let me explain the current situation:
- I don't see the benefit of blocking the new feature (UASTs in FilePairs) till we fix a problem that we do not have yet (nobody is using PostgreSQL yet)
- If we don't block this PR, it would be merged, so the PostgreSQL support would be broken the facto. If we leave the docs (and the code) as they currently are, it could confuse people trying to use PostgreSQL from now on.
- I don't see the benefit of releasing something we already know it is broken.
- I'm not purposing to drop definitively the PostgreSQL support; It can be resumed considering the issue https://github.com/src-d/code-annotation/issues/222 . The commit dropping PostgreSQL support 94d41ab can be reverted; nothing is lost.

carlosms:
I agree with @bzz and @smacker. The way I see it removing all the postgres code is too drastic. And if we decide to remove it, I'd prefer to see it in a new PR of its own.

dpordomingo:
Could you @bzz @smacker @carlosms review the PR again considering the comments below?

I Rebased and restored the offending commit dropping PostgreSQL support as you suggested. I found some other [problems with PostgreSQL](https://github.com/src-d/code-annotation/milestone/1) that should be solved separately.

I rebased over master, and fixed some other problems that I found in the original branch:
- 7f05627 Avoid star select to avoid problems with some sources
- edf2bbb Reorder columns to ensure that export command works

I also added support for BLOB data type if PostgreSQL DB is used. I tried it locally, and I could only import the first 100 FilePairs, that were properly processed and its content was the same that when they are imported into an SQLite database. The rest 1900 FilePairs could not be imported because of an unrelated problem as described by https://github.com/src-d/code-annotation/issues/224
- 836d8db Add PostgreSQL support for BLOB

cli/import/examples/example.sql
dpordomingo: done by:
- 836d8db Add PostgreSQL support for BLOB

dpordomingo (approved):
Could you @bzz @smacker @carlosms review the PR again considering the comments below?

I Rebased and restored the offending commit dropping PostgreSQL support as you suggested. I found some other [problems with PostgreSQL](https://github.com/src-d/code-annotation/milestone/1) that should be solved separately.

I rebased over master, and fixed some other problems that I found in the original branch:
- 7f05627 Avoid star select to avoid problems with some sources
- edf2bbb Reorder columns to ensure that export command works

I also added support for BLOB data type if PostgreSQL DB is used. I tried it locally, and I could only import the first 100 FilePairs, that were properly processed and its content was the same that when they are imported into an SQLite database. The rest 1900 FilePairs could not be imported because of an unrelated problem as described by https://github.com/src-d/code-annotation/issues/224
- 836d8db Add PostgreSQL support for BLOB

carlosms (approved):

bzz (commented):
LGTM

Thank you @dpordomingo for taking care, addressing all feedback. 👍 for having tests

bzz (approved):

PR #211, opened 2018-03-12
Show new lines, tabs, whitespaces

smacker:
Fixes: https://github.com/src-d/code-annotation/issues/176

screen shot 2018-03-12 at 12 23 45

### DIFF
--- a/server/handler/file_pairs.go
+++ b/server/handler/file_pairs.go
@@ -33,11 +33,18 @@ func GetFilePairDetails(repo *repository.FilePairs, diff *service.Diff) RequestP
 			return nil, serializer.NewHTTPError(http.StatusNotFound, "no file-pair found")
 		}
 
+		var preprocessors []service.DiffPreprocessorFunc
+
+		if r.URL.Query().Get("showInvisible") == "1" {
+			preprocessors = append(preprocessors, service.ReplaceInvisible)
+		}
+
 		diffString, err := diff.Generate(
 			filePair.Left.Path,
 			filePair.Right.Path,
 			filePair.Left.Content,
 			filePair.Right.Content,
+			preprocessors...,
 		)
 		if err != nil {
 			return nil, err
@@ -45,9 +52,6 @@ func GetFilePairDetails(repo *repository.FilePairs, diff *service.Diff) RequestP
 
 		leftLOC := len(strings.Split(filePair.Left.Content, "\n"))
 		rightLOC := len(strings.Split(filePair.Right.Content, "\n"))
-		if err != nil {
-			return nil, err
-		}
 
 		return serializer.NewFilePairResponse(filePair, diffString, leftLOC, rightLOC), nil
 	}
--- a/server/service/diff.go
+++ b/server/service/diff.go
@@ -1,6 +1,10 @@
 package service
 
-import "github.com/pmezard/go-difflib/difflib"
+import (
+	"strings"
+
+	"github.com/pmezard/go-difflib/difflib"
+)
 
 // Diff service generates diff for files
 type Diff struct {
@@ -12,8 +16,20 @@ func NewDiff() *Diff {
 	return &Diff{context: 6} // keep it hard coded for now
 }
 
+// DiffPreprocessorFunc type is function signature to preprocess diffs
+type DiffPreprocessorFunc func(string) string
+
 // Generate return unified diff string for 2 files
-func (d *Diff) Generate(nameA, nameB, contentA, contentB string) (string, error) {
+func (d *Diff) Generate(nameA, nameB, contentA, contentB string, preprocessors ...DiffPreprocessorFunc) (string, error) {
+	for _, p := range preprocessors {
+		contentA = p(contentA)
+		contentB = p(contentB)
+	}
+
+	return d.generate(nameA, nameB, contentA, contentB)
+}
+
+func (d *Diff) generate(nameA, nameB, contentA, contentB string) (string, error) {
 	diff := difflib.UnifiedDiff{
 		A:        difflib.SplitLines(contentA),
 		B:        difflib.SplitLines(contentB),
@@ -23,4 +39,12 @@ func (d *Diff) Generate(nameA, nameB, contentA, contentB string) (string, error)
 	}
 
 	return difflib.GetUnifiedDiffString(diff)
+}
+
+// ReplaceInvisible preprocessor function that replace invisible character with visible onces
+func ReplaceInvisible(content string) string {
+	content = strings.Replace(content, " ", "·", -1)
+	content = strings.Replace(content, "\t", "→", -1)
+	content = strings.Replace(content, "\r", "^M", -1)
+	return strings.Replace(content, "\n", "↵\n", -1)
 }
--- a/server/service/diff_test.go
+++ b/server/service/diff_test.go
@@ -39,6 +39,20 @@ func (suite *DiffSuite) TestDiff() {
 	assert.Equal(ac, acDiff)
 }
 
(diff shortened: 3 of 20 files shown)


ricardobaeta:
Awesome @smacker !

server/handler/file_pairs.go
```diff
@@ -33,11 +33,18 @@ func GetFilePairDetails(repo *repository.FilePairs, diff *service.Diff) RequestP
...
+		}
+
 		diffString, err := diff.Generate(
 			filePair.Left.Path,
 			filePair.Right.Path,
 			filePair.Left.Content,
 			filePair.Right.Content,
+			preprocessors...,
 		)
 		if err != nil {
```
dpordomingo, line 49: This one is duplicated some lines below https://github.com/src-d/code-annotation/pull/211/files#diff-fcf9e14a0be9f2c9ac7cd962eaa11d10R55

server/service/diff.go
```diff
@@ -12,8 +16,25 @@ func NewDiff() *Diff {
...
+
 // Generate return unified diff string for 2 files
-func (d *Diff) Generate(nameA, nameB, contentA, contentB string) (string, error) {
+func (d *Diff) Generate(nameA, nameB, contentA, contentB string, preprocessors ...DiffPreprocessorFunc) (string, error) {
+	for _, p := range preprocessors {
+		contentA = p(contentA)
+		contentB = p(contentB)
+	}
+
+	diffString, err := d.generate(nameA, nameB, contentA, contentB)
```
dpordomingo: this line could be replaced by 
```go
retun d.generate(nameA, nameB, contentA, contentB)
```
And then the code below that would be unnecessary

server/service/diff.go
```diff
@@ -24,3 +45,16 @@ func (d *Diff) Generate(nameA, nameB, contentA, contentB string) (string, error)
 
 	return difflib.GetUnifiedDiffString(diff)
 }
+
+// DiffReplaceInvisible preprocessor function that replace invisible character with visible onces
+func DiffReplaceInvisible(content string) string {
+	lines := strings.Split(content, "\n")
```
dpordomingo: Do we need to split &rarr; replace &rarr; join?

Could we replace it by...
```go

func ReplaceInvisible(content string) string {
	line := strings.Replace(content, " ", "·", -1)
	line = strings.Replace(line, "\t", "→", -1)
	line = strings.Replace(line, "\r", "^M", -1)
	line = strings.Replace(line, "\n", "↵\n", -1)
	return line	
}
```

server/service/diff.go
```diff
@@ -24,3 +45,16 @@ func (d *Diff) Generate(nameA, nameB, contentA, contentB string) (string, error)
 
 	return difflib.GetUnifiedDiffString(diff)
 }
+
+// DiffReplaceInvisible preprocessor function that replace invisible character with visible onces
+func DiffReplaceInvisible(content string) string {
```
dpordomingo: `Diff` should not be part of the function name

server/service/diff_test.go
```diff
@@ -39,6 +39,17 @@ func (suite *DiffSuite) TestDiff() {
...
 }
 
+func (suite *DiffSuite) TestDiffReplaceInvisible() {
+	assert := suite.Assert()
+	diffStr, err := readFile("./testdata/diff_with_invisible.diff")
+	assert.NoError(err)
+
+	assert.Equal(
+		service.DiffReplaceInvisible(diffStr),
+		"---·a.txt^M↵\n+++·b.txt^M↵\nline^M↵\n····tabbed·line^M↵\n····spaced·line^M↵\n",
```
dpordomingo: since `Diff` should not be part of the `ReplaceInvisible` function name, it should be applied everywhere, what means that the test data should not be a diff file (because then, it looks a different thing)

dpordomingo: If the test input data is loaded from a file, the expected data should be also loaded from a file.

imo, both: input and expected should be defined inside of the test case function instead of inside an external file (they're too simple to be hidden inside a file), what is also more **explicit and clear**.
```go
input := "line\r\n".
  "\ttabbed line\n".
  "    spaced line\n"
expected := "line^M↵\n".
    "→tabbed·line↵\n".
    "····spaced·line↵\n"
assert := suite.Assert().Equal(input, expected)
```

server/service/testdata/diff_with_invisible.diff
```diff
@@ -0,0 +1,5 @@
+--- a.txt
```
dpordomingo: if input/expected is inside of the test case, this should be not necessary.

server/handler/file_pairs.go
```diff
@@ -33,11 +33,18 @@ func GetFilePairDetails(repo *repository.FilePairs, diff *service.Diff) RequestP
 			return nil, serializer.NewHTTPError(http.StatusNotFound, "no file-pair found")
 		}
 
+		var preprocessors []service.DiffPreprocessorFunc
+
+		if r.URL.Query().Get("showInvisible") != "" {
```
dpordomingo: for consistency with [front api call](https://github.com/src-d/code-annotation/pull/211/files#diff-284817f1b6aad47c0ac6d61bd72f474cR103), it should be:
```go
		if r.URL.Query().Get("showInvisible") == "1" {
```

dpordomingo (changes requested):
thanks!
storing the user preference in the local storage was a great idea, and it works like a charm!

I'd only require some cosmetic and organizational changes.

server/service/diff.go
```diff
@@ -24,3 +40,11 @@ func (d *Diff) Generate(nameA, nameB, contentA, contentB string) (string, error)
 
 	return difflib.GetUnifiedDiffString(diff)
 }
+
+// ReplaceInvisible preprocessor function that replace invisible character with visible onces
+func ReplaceInvisible(content string) string {
+	content = strings.Replace(content, " ", "·", -1)
+	content = strings.Replace(content, "\t", "→", -1)
```
carlosms, line 47: How about replacing tab with ```'→   '``` to try to keep indentation? I think a single char indentation may make code less readable.

carlosms (approved):
Looks good to me, I just left a suggestion for the tab character.

server/handler/file_pairs.go
```diff
@@ -35,8 +35,8 @@ func GetFilePairDetails(repo *repository.FilePairs, diff *service.Diff) RequestP
 
 		var preprocessors []service.DiffPreprocessorFunc
 
-		if r.URL.Query().Get("showInvisible") != "" {
-			preprocessors = append(preprocessors, service.DiffReplaceInvisible)
+		if r.URL.Query().Get("showInvisible") != "1" {
```
dpordomingo: ups! it should be `==`
:wink:

dpordomingo (changes requested):

server/service/diff.go
dpordomingo: I'm used to seeing only one &rarr; symbol.
If you want more space, we could use &#10230; (`&#10230;` or `&#x27f6;`).
but whatever you prefer.

carlosms: That's odd, I kind of expect any editor to show the symbol _and_ indent. Like VS Code for example:
tabs

server/handler/file_pairs.go
smacker: ahahahaha. 🤦‍♂️ 

fixed.

server/service/diff.go
smacker: I definitely saw *only* → many times.

dpordomingo (approved):
LGTM
If @carlosms has not an strong opinión about indentation chars, please squash and merge

bzz (approved):

Issue #215, opened 2018-03-14
Prepare and deploy Production environment


dpordomingo:
In order to have a Production environment we need:
- Retrieve data:
  - [x] to know the HOSTNAME to be used for production,
  - [x] to create and communicate the new `GA_TRACKING_ID` for testing purposes &rarr; https://github.com/src-d/code-annotation/issues/216
  - [x] create a new GitHub OAuth application for prod, using Prod callback &rarr; https://github.com/src-d/code-annotation/issues/217
  - [x] to communicate to @src-d/infrastructure the secrets of the new GitHub OAuth application
- Prepare the environment:
  - [x] `.drone.yml` and helm charts &rarr; https://github.com/src-d/code-annotation/pull/218
- Deploy
  - [x] configure Kubernetes secrets &rarr; https://github.com/src-d/issues-infrastructure/issues/152
  - [x] stop the annotation process being done in Staging, &rarr; https://github.com/src-d/code-annotation/pull/237
  - [x] deploy CAT into Staging &rarr; unauthorizing all no @src-d/applications members,
  - [x] deploy CAT into Prod
- Create the Production internal DB 
  - [x] download, modify and upload the new version of the internal DB into Prod and Staging &rarr; https://github.com/src-d/code-annotation/issues/214
- Test and celebrate:
  - [x] configure GA pannels &rarr; https://github.com/src-d/code-annotation/issues/241
  - [x] notify CAT users that from now on it should be used the Production env,

rporres:
hostname: **code-annotation.srcd.run**
ingress.globalStaticIpName: code-annotation-production

Helm chart is the same, only the values passed change

(hostname edited by @dpordomingo as agreed by https://github.com/src-d/code-annotation/issues/217#issuecomment-378615285)

PR #218, opened 2018-03-15
Configure the Production environment

dpordomingo:
required by https://github.com/src-d/code-annotation/issues/215

This PR prepares the Prod env.

Changes done:
- this application now is deployed into Staging and Production following [our continuous-delivery guide](https://github.com/src-d/guide/blob/master/engineering/continuous-delivery.md),
  - every new tag will trigger a deploy into Production,
- `gaTrackingID` must be passed by drone.io when deploying
  - Staging will have its own `GA_TRACKING_ID` for testing purposes
- Staging will no longer be accessible for no @src-d/applications members

Other changes that should be done by @src-d/infrastructure just after merging this PR:
- https://github.com/src-d/issues-infrastructure/issues/152

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -2,7 +2,7 @@ workspace:
   base: /go
   path: src/github.com/src-d/code-annotation
 
-branches: [master, staging, release/*]
+branches: [master, staging]
 
 # ANCHORS
 
@@ -27,6 +27,14 @@ docker_image: &docker_image
   dockerfile: Dockerfile
   debug: true
 
+helm_deploy: &helm_deploy
+  image: quay.io/ipedrazas/drone-helm
+  skip_tls_verify: true
+  chart: ./helm-charts/code-annotation
+  release: code-annotation
+  tiller_ns: kube-system
+  wait: true
+
 # PIPELINE STEPS
 
 pipeline:
@@ -36,7 +44,7 @@ pipeline:
     debug: true
 
 
-  # deployment to staging environment
+  # deployment to staging environment when staging is pushed
 
   build_stg:
     <<: *build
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -145,6 +145,10 @@ CAT_OAUTH_RESTRICT_ACCESS=org:my-organization
 CAT_OAUTH_RESTRICT_REQUESTER_ACCESS=team:123456
 ```
 
+## Deployment
+
+This site is deployed in `production` and in `staging` following our [web application deployment workflow](https://github.com/src-d/guide/blob/master/engineering/continuous-delivery.md)
```
bzz, line 150: This is something important - we should think (and document it as well) that it's not a web site or a service, it's an application. This has consequences on many levels, including the way structure the release, package&distribute artifacts, monitor it, etc.

Ofcourse as a web application, it can be deployed by different people in different scenarios on different environments (and that's what we aiming for), but I think it's worth highlighting it here and making a clear distinction between general expectations and sourced-specific things.

How about 
 - changing the title from `Deployment` -> `source{d} internal deployment`
 - changing `site` -> `application`
 - `production and staging  sourced{d} environments`

bzz (approved):
👍  LGTM, sans minor doc issue, highlighted above.

README.md
dpordomingo: I just copied the code from Blog and Landing but yes: your proposals are even better and they could be also applied there.
I'll proceed with your suggestions before merging, Thanks!

carlosms (approved):

helm-charts/code-annotation/templates/deployment.yaml
```diff
@@ -50,7 +50,7 @@ spec:
                   name: {{ $secretName  }}
                   key: jwt_signing_key
             - name: CAT_GA_TRACKING_ID
-              value: "{{ .Values.deployment.gaTrackingID }}"
+              value: "{{ required "gaTrackingID is required" .Values.deployment.gaTrackingID }}"
```
dpordomingo, line 53: I found two different ways to write it:

wrapping the value [with quotes](https://github.com/src-d/code-annotation/pull/218/files#diff-f03305ba93fd99a4f9a9c72bcb7d5d06R26):
```
image: "{{ .Values.image.repository }}:{{ required "Image tag is required" .Values.image.tag }}"
```

leaving the value [without wrapper quotes](https://github.com/dpordomingo/code-annotation/blob/93a4e4259bd4d9c1684a47af0cc1a2bc1cf70c5d/helm-charts/code-annotation/templates/ingress.yaml#L20):
```
- host: {{ required "Hostname is missing" .Values.ingress.hostname }}
```

@rporres should I use one style for this `CAT_GA_TRACKING_ID` env value?

dpordomingo:
@bzz your documentation suggestions were added by https://github.com/src-d/code-annotation/pull/218/commits/14b4c49909aedb3c03332db12ee7d72981804058
Many thanks for proposing them!

helm-charts/code-annotation/templates/deployment.yaml
rporres: All env variables must be strings, so they must be quoted to avoid YAML interpreting as numbers

.drone.yml
```diff
@@ -2,7 +2,7 @@ workspace:
   base: /go
   path: src/github.com/src-d/code-annotation
 
-branches: [master, staging, release/*]
+branches: [master, staging]
```
rporres, line 5: why have you removed `release/*`?

rporres (commented):
couple of minor questions, looks good

sorry for the delay, I've been busy these days

.drone.yml
dpordomingo: It was copied from Landing `.drone.yml`, that **run tests**, and triggers the deploy;
In Landing (and blog and talks), tests are also run on `release/*` branches.
If we want to reproduce :point_up: this behavior, **it should be done in `.travis.yml` file**
 (because in CAT tests are run from Travis instead of from Drone) but it is out of the scope of this PR (and we didn't consider it yet for this app)

Does it answer your question @rporres ?

rporres: sure! :)

rporres (approved):

PR #232, opened 2018-03-27
Remove diff column

dpordomingo:
Diff was no longer needed since https://github.com/src-d/code-annotation/pull/194

The Staging internal database will be migrated with (via https://github.com/src-d/code-annotation/pull/221)
```shel
go run cli/migrations/*.go diff-rm-col internal.db
```


carlosms (approved):
:+1:

smacker (approved):

Issue #198, opened 2018-03-02
Integrate reviewer \w REST API


bzz:
As soon as there is a 
 - [ ] REST API for feature extraction https://github.com/src-d/backlog/issues/1202
 - [x] UASTs are stored in DB ~https://github.com/src-d/code-annotation/issues/197~ &rarr; https://github.com/src-d/code-annotation/issues/214

we would need to call an instance of this REST API through a configurable URL.

REST API input:
 - 2 UASTs (optional)

REST API output:
 - data that need to be shown in reviewer UI

Preliminary suggested by @vmarkovtsev JSON in details
```json
[
{
    "tab": "first scheme",
    "object1": [
        {"feature": "name1", "type": "number", "value": 100500},
        {"feature": "name2", "type": "hash", "value": "abcdef0123456789"}
    ],
    "object2": [
        {"feature": "name1", "type": "number", "value": 100501},
        {"feature": "name2", "type": "hash", "value": "abcdef1234567890"}
    ],
    "pair": [
        {"feature": "score", "type": "number", "value": 0.95}
    ],
}, {
    "tab": "second scheme",
    "object1": [
        {"feature": "name1", "type": "number", "value": 100500},
        {"feature": "name2", "type": "hash", "value": "abcdef0123456789"}
    ],
    "object2": [
        {"feature": "name1", "type": "number", "value": 100501},
        {"feature": "name2", "type": "hash", "value": "abcdef1234567890"}
    ],
    "pair": [
        {"feature": "score", "type": "number", "value": 0.95}
    ],
}
]
```

but @mcuadros suggested using a simple structure, that as much as possible can be just "shown" or printed in the reviewer UI.

Part of this issue is to determine this simplest-possible, but extensible format.

dpordomingo:
As talked irl with @bzz UAST will be an optional data in the CAT internal database

dpordomingo:
From our meetings we understood that:
- the payload sent to the API will be the two UAST stored in the internal DB for the FilePair,
- the current UI would be kept if the data to represent fits on it,
- if the data to represent does not fit the current UI, it will be printed as it comes from the API,

Considering the assertions above, in order to use the current UI, the API Response schema should be:

```go
type FilePairFeatures {
    Object1 []Feature
    Object2 []Feature
    Pair Feature
}

type Feature {
    Name string
    Value float32
}
```

This schema fits the proposed by @vadim but has the following modifications:
- A `Feature` has not type &rarr; it is always a `float` number,
[...]

PR #233, opened 2018-03-28
Request features from Review section with one call

### LINKED ISSUE #229: Reviewer panel is not fetching features when reviewed FilePair changes
Reviewer panel is not fetching features when reviewed FilePair changes 

The url is not changing neither

image

image

dpordomingo:
required by #198 and https://github.com/src-d/backlog/issues/1202
fixes #229

As defined by https://github.com/src-d/code-annotation/issues/198#issuecomment-375703060, the [ML Feature Extractor API](https://github.com/src-d/backlog/issues/1202) will take a couple of UAST, and it will return a bunch of features per each UAST, and a `score` as a _pair-feature_

This PR prepares the frontend to do one call per FilePair being reviewed to obtain the features of each file, and the score itself.
This PR prepares the backend to answer following the expected schema.

No UI changes were done.

It was also added a 
3445eb0 commit to fix #229 because otherwise the FilePairs beyond the first one cannot be reviewed.

## Pending work

To be done once the [ML Feature Extractor API](https://github.com/src-d/backlog/issues/1202) is ready, but that does not block this PR:
- the new API musth be called from the features handler (pointed by the [TODO](https://github.com/src-d/code-annotation/pull/233/files#diff-ae70b50bc82721e179675b32e8aa5845R34)),
- the [`model.Features`](https://github.com/src-d/code-annotation/blob/master/server/model/models.go#L64) must be removed,
- the [`features` db table](https://github.com/src-d/code-annotation/blob/master/server/dbutil/dbutil.go#L65) must be removed &rarr; running `features-drop-table` migration from https://github.com/src-d/code-annotation/pull/221

### DIFF
--- a/server/handler/features.go
+++ b/server/handler/features.go
@@ -3,24 +3,51 @@ package handler
 import (
 	"net/http"
 
-	"github.com/go-chi/chi"
+	"github.com/src-d/code-annotation/server/model"
 	"github.com/src-d/code-annotation/server/repository"
 	"github.com/src-d/code-annotation/server/serializer"
 )
 
 // GetFeatures returns a function that returns a *serializer.Response
 // with the list of features for blobId
-func GetFeatures(repo *repository.Features) RequestProcessFunc {
+func GetFeatures(filePairRepo *repository.FilePairs, featuresRepo *repository.Features) RequestProcessFunc {
 	return func(r *http.Request) (*serializer.Response, error) {
-		blobID := chi.URLParam(r, "blobId")
-
-		// in the future it should take file by blobID from DB
-		// and make API request to ML system
-		features, err := repo.GetAll(blobID)
+		filePairID, err := urlParamInt(r, "pairId")
 		if err != nil {
 			return nil, err
 		}
 
-		return serializer.NewFeaturesResponse(features), nil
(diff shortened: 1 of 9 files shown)


carlosms (approved):
:+1:

smacker:
I didn't get why we need to change it to 1 call now.
As I understand there is no API yet. And we can actually use https://github.com/src-d/gemini/blob/master/src/main/python/server.py as such api. It accepts 1 uast + params for extraction and returns list of features.

I would better extract fix for #229 in a separate commit and wait with other changes as api isn't ready yet.

on another hand changes are small, so if you want to merge it, it's okay.

dpordomingo:
@smacker all your doubts are properly explained in the PR description; for example:
- we need to call the API with two UAST bec it will return the similarity score between them,
>_[...] the [ML Feat. Extr. API](https://github.com/src-d/backlog/issues/1202) will take a couple of UAST,and it will return [...] a `score` as a **pair-feature**_

bzz (approved):

bzz (approved):
Although https://github.com/src-d/code-annotation/pull/233#issuecomment-378187580 has the point, the explanation about necessity of this work holds and even if it's a bit preliminary, it LGTM.

PR #235, opened 2018-04-03
Add a favicon

dpordomingo:
The browser automatically requests a `/favicon.ico` on every request;
since we have no one, the server is returning the `/index.html` what is wrong.

image

I added it using the current app logo
image


dpordomingo:
@bzz I think we could consider this a hotfix (to avoid the none-favicon), and just merge it.
In case it is needed a new favicon, it could be pushed in a new PR.

bzz (approved):

Issue #217, opened 2018-03-15
Create a new GitHub OAuth App for Prod env


dpordomingo:
As required by [Prepare and deploy Production environment #215](https://github.com/src-d/code-annotation/issues/215) it is needed to create a new GitHub OAuth App with the following data:

- Application name: `code-annotation`
- Homepage: ~~`code-annotation.sourced.tech`~~ &rarr; `http://code-annotation.srcd.run`
- Authorization callback URL: `http://code-annotation.srcd.run/oauth-callback`

The secrets should be communicated to @src-d/infrastructure team

mcuadros:
This application never should be exposed under `sourced.tech` domain since is an internal tool

dpordomingo:
The proposed hostname was proposed by @rporres when I asked him for going to Prod; from my point of view, there is no necessity on having exactly that hostname, so it could be whatever @mcuadros / @rporres / @bzz purposes.
Since the staging one is `code-annotation-staging.srcd.run`
what about if the prod one would be `code-annotation-production.srcd.run`? would you like it?

rporres:
All production cluster applications are exposed using `sourced.tech`, hence my assumption.

`code-annotation-production.srcd.run` looks fine to me.

dpordomingo:
Since @rporres agrees too, I updated the issue desc with the proposed hostname.
If you have any other proposal @mcuadros , I have no concern about it, so feel free to use it.

mcuadros:
I just created the OAuth, for the URL : `http://code-annotation.srcd.run/`

dpordomingo:
keys provided by slack; thanks

PR #236, opened 2018-04-04
Prepare Production, and make Staging private

dpordomingo:
After preparing Production environment https://github.com/src-d/code-annotation/pull/218 , some other changes must be done to:

- 196230c Convert Staging to a real testing env, making it accessible for [Applications members](https://github.com/orgs/src-d/teams/applications)
- 307509c Use production url as defined by https://github.com/src-d/code-annotation/issues/217#issuecomment-378615285

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -97,6 +97,6 @@ pipeline:
     <<: *helm_deploy
     prefix: PROD
     secrets: [ PROD_API_SERVER, PROD_KUBERNETES_TOKEN ]
-    values: ingress.globalStaticIpName=code-annotation-production,ingress.hostname=code-annotation.sourced.tech,image.tag=${DRONE_TAG},deployment.gaTrackingID=UA-109494282-2,secretName=code-annotation-drone-workaround
+    values: ingress.globalStaticIpName=code-annotation-production,ingress.hostname=code-annotation.srcd.run,image.tag=${DRONE_TAG},deployment.gaTrackingID=UA-109494282-2,secretName=code-annotation-drone-workaround
     when:
       event: [tag]
--- a/.drone.yml
+++ b/.drone.yml
@@ -64,7 +64,7 @@ pipeline:
     <<: *helm_deploy
     prefix: STG
     secrets: [ STG_API_SERVER, STG_KUBERNETES_TOKEN ]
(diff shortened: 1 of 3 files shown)


.drone.yml
```diff
@@ -97,6 +97,6 @@ pipeline:
     <<: *helm_deploy
     prefix: PROD
     secrets: [ PROD_API_SERVER, PROD_KUBERNETES_TOKEN ]
-    values: ingress.globalStaticIpName=code-annotation-production,ingress.hostname=code-annotation.sourced.tech,image.tag=${DRONE_TAG},deployment.gaTrackingID=UA-109494282-2,secretName=code-annotation-drone-workaround
+    values: ingress.globalStaticIpName=code-annotation-production,ingress.hostname=code-annotation.srcd.run,image.tag=${DRONE_TAG},deployment.gaTrackingID=UA-109494282-2,secretName=code-annotation-drone-workaround
```
dpordomingo, line 100: is this conf compatible @rporres ?
```
ingress.globalStaticIpName=code-annotation-production
ingress.hostname=code-annotation.srcd.run
```
I'm not sure if we should use a new `globalStaticIpName` suffixed as `-production` while we're having a `hostname` following the `staging` pattern.

Is this something I should care about?
Does it affect the place where you store the secrets and these configurations?
(most probably I'm mixing concepts, so please ping me when you have time to explain it to me irl :dancer: )

dpordomingo (commented):
@rporres PTAL

bzz (approved):
LGTM. Let's wait for @rporres feedback as well.

Only App team members will be able to login to staging now.

I guess we just need to make sure that on production, same users as before would be able to log in, and see their previous answers.

dpordomingo:
Since the staging DB will be copied to production, all workers will keep their previous work already done in Staging.
The Production ACL will be the same as currently in Staging: `restrictAccessGroup:org:src-d`, so nothing changed about it in Prod.

Does it answer your questions @bzz ?

.drone.yml
rporres: Please use `code-annotation-production.srcd.run` as hostname. The rest is ok

rporres (changes requested):

.drone.yml
dpordomingo: But that's the hostname defined by @mcuadros in the GH OAuth App as he wrote at https://github.com/src-d/code-annotation/issues/217#issuecomment-378615285

Should we ask him to change it?

rporres: d'oh, ok. I don't like that name, but I guess it's late to change it. I will change it on my side...

rporres (approved):

PR #245, opened 2018-10-10
added helm node selector

rporres:
Signed-off-by: Rafael Porres Molina <rafa@sourced.tech>

PR #249, opened 2018-10-17
Revert "added helm node selector"

meyskens:
Reverts src-d/code-annotation#245

The values will cause an error on deploy.


rporres (approved):
Just for the record, the problem is drone-helm escaping quotes it should not

PR #250, opened 2018-10-17
use Helm values files

meyskens:
Signed-off-by: Maartje Eyskens <maartje@eyskens.me>
Is like #245 but using values files since drone-helm escapes quotes incorrect.


rporres (approved):
Once merge conflicts are solved, LGTM :)

carlosms (approved):
:+1:

